PR #2, opened 2015-03-24
WEBSITE-272,WEBSITE-273 CSS fixes

DavideD:
A patch for [WEBSITE-272](https://hibernate.atlassian.net/browse/WEBSITE-272) and [WEBSITE-273](https://hibernate.atlassian.net/browse/WEBSITE-273)

For the tags, I've put everything in one column like the author list. The reason is that the tag names are quite long and therefore it looks better this way.

Let me know what you think

### DIFF
--- a/stylesheets/styles.scss
+++ b/stylesheets/styles.scss
@@ -354,6 +354,16 @@ footer.container {
   margin-bottom: 1em;
 }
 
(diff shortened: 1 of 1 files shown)


hferentschik:
Merged thanks

PR #3, opened 2015-03-26
WEBSITE-274 Include disqus using a template

DavideD:
https://hibernate.atlassian.net/browse/WEBSITE-274

This patch will also update the posts folder.
 I tried to add a counter with the comments in the home page, but everything I tried didn't work.

I'm not sure we can use the awestruct extension since the disqus_url used is different.
I need to talk with @hferentschik  about it


DavideD:
> Is the slug enough or should be use a prefix of the creation date as well?

With the addition of the date we have a better identifier.
Using the identifier seems to be the suggested way since it is not affected by the website domain and it helps in case we use different domains.

@hferentschik I wrote here what we discussed in IRC in case someone else read this pull request

hferentschik:
> I wrote here what we discussed in IRC in case someone else read this pull request

+1

hferentschik:
merged, thanks

PR #5, opened 2015-03-27
WEBSITE-280 tags

DavideD:
https://hibernate.atlassian.net/browse/WEBSITE-280

This should make the tags cloud nicer

### DIFF
--- a/stylesheets/styles.scss
+++ b/stylesheets/styles.scss
@@ -364,7 +364,7 @@ footer.container {
   margin-top: 0;
 }
 
-.tags-0 {
+.tags-0, .tags-1 {
   display: none;
 }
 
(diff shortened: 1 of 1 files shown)


hferentschik:
Apart from the fact that I don't quite understand how a bit of css can make this happen, I am also a bit surprised by the result. I like that is it now an actual tag cloud, but how does it work. For example, the quite important tags of 'Hibernate Valdiator' and 'Hibernate OGM'. Both are there in the plain tag list, but are gone in the cloud, even though they have more than one post against them and the tag url actually exists and I can browse the posts under this url. Feature? Bug?

hferentschik:
Closing this pull request, since it seems to be superseded by #6

DavideD:
It's not a bug, I'm applying "display: none" to all the span elements with the class "tags-0" or "tags-1", making them disappear.

The error is that I misunderstood the meaning of tags-0 and tags-1, I thought it was connected to the number of posts but it's probably the ones with less posts compare to the other tags.

PR #7, opened 2015-03-30
WEBSITE-284 Display "tags-1" tags

DavideD:
This pull request will re-enable the tag link with class tags-1, showing again Hibernate OGM and Validator.

The problem is that a lot of tags will reappear again.

@emmanuelbernard this is related to this comment from @hferentschik:

>Apart from the fact that I don't quite understand how a bit of css can make this happen, I am also a bit surprised by the result. I like that is it now an actual tag cloud, but how does it work. For example, the quite important tags of 'Hibernate Valdiator' and 'Hibernate OGM'. Both are there in the plain tag list, but are gone in the cloud, even though they have more than one post against them and the tag url actually exists and I can browse the posts under this url. Feature? Bug?

### DIFF
--- a/stylesheets/styles.scss
+++ b/stylesheets/styles.scss
@@ -380,9 +380,13 @@ footer.container {
   display: inline-block;
 }
 
-// Disable tags with less than 2 posts
-.tags-cloud .tags-0, .tags-cloud .tags-1 {
+.tags-cloud .tags-0 {
   display: none;
+}
+
+.tags-cloud .tags-1 {
+     font-size: 1.15em;
+     opacity: 0.85;
 }
 
 .tags-cloud .tags-2 {


hferentschik:
Is there anyway to manually influence the tag cloud extension, eg can one explicitly include/exclude tags? Does it make sense to do a tag cleanup? If we decide on a list of "un-wanted" tags we can already exclude them as part of the erb file creation!?

DavideD:
> can one explicitly include/exclude tags?

You can for sure not render a link, I'm not sure if there is an easy way to filter tags. I will check.

> Does it make sense to do a tag cleanup? 

Yes, some tags don't seem very useful and some others need to be merged. For example:
* Hibernate OGM, OGM -> Hibernate OGM
* JBoss Tools Eclipse, JBoss Tools -> JBoss Tools

I'm sure there are others that can be merged and some that can be removed

> If we decide on a list of "un-wanted" tags we can already exclude them as part of the erb file creation!?

I guess we can, I suppose we also want some validation for new posts.

hferentschik:
> Yes, some tags don't seem very useful and some others need to be merged. For example: ...

Right, I guess a good start would be to complete the list of tags to be deleted/merged. I created WEBSITE-299 for this. 

> I guess we can, I suppose we also want some validation for new posts.

Good point, if we want ongoing validation we could add a helper function in awestruct. Maybe a YAML file to configure it all. See WEBSITE-299

WDYT?

gunnarmorling:
@DavideD, just seeing that this one is still open. What should happen about it?

DavideD:
I think it should be included.
The idea was to solve the issues related to the cloud, but it hides too many tags.

I will solve the issue with [WEBSITE-299](https://hibernate.atlassian.net/browse/WEBSITE-299)

DavideD:
I will solve the issue with https://github.com/hibernate/in.relation.to/pull/11

PR #8, opened 2015-03-30
WEBSITE-281 Mobile fixes

DavideD:
https://hibernate.atlassian.net/browse/WEBSITE-281

PR #9, opened 2015-03-30
WEBSITE-290 Add Home And Feed in single post page

DavideD:
https://hibernate.atlassian.net/browse/WEBSITE-290

### DIFF
--- a/_layouts/blog/blog-post.html.haml
+++ b/_layouts/blog/blog-post.html.haml
@@ -15,7 +15,25 @@ layout: project
           %i.icon-circle.icon-stack-base.icon-fixed-width
           %i{:class => "icon icon-light icon-fixed-width icon-quote-right"}
           /icon-rss icon-signal
-      %hr.divider
+    %hr.divider
+    %ul.nav.nav-list
+      -# removes index.html and then replace download.html to download/
+      - current_path = page.output_path.sub(/index\.html$/, "").sub(/\.html$/,"/")
+
+      - href = "/"
+      - active = false
+      %li{:class => "#{(active ? "active" : "")}"}
+        %a{:href => "#{relative("#{href}")}"}
+          %i{:class => "icon-home menu-icon"}
+          Home
+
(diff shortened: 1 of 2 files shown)


DavideD:
Rebased on top of https://github.com/hibernate/in.relation.to/pull/8
added https://hibernate.atlassian.net/browse/WEBSITE-294

DavideD:
I made the title of the posts a bit darker for WEBSITE-294, I don't know if anyone else will like it  :)
Feel free to suggest a different color.

hferentschik:
A couple of thoughts. The links appear now also on single post pages, so that is good. I am wondering though about the "Home" link. Starting on hibernate.org the little house icon is used for a so called "About" link which is context sensitive to the selected project at the top. Would it not make sense to call it "About" as will on in.relation.to? It would then point indeed to the root url of in.relation.to. Using "home" implies to me hibernate.org. Using "About" makes maybe also more sense when we have a hero text box (WEBSITE-279). Alternatively I could live with removing the link. 

> I made the title of the posts a bit darker for WEBSITE-294, I don't know if anyone else will like it :)
Feel free to suggest a different color.

Personally I would use the same color (the greyish) which is used to render the title when a single post is displayed. I guess blue indicates a link, but I still feel it is too hard to distinguish the separate blogs on the landing page. Actually in a perfect world, I would not make subsection headings linkable on the landing page (or on any page where multiple posts are displayed). Not sure whether that is possible.

DavideD:
> Would it not make sense to call it "About" as will on in.relation.to

Maybe in the contest of the blog it make more sense to call it "News" or "Latest" or something like that.

> Personally I would use the same color (the greyish) 

I will try, if you want to have a look before I send a pull request, this is the code for the color: #656565

DavideD:
I've applied the color change to the title of the post

hferentschik:
Merged. I think not having blue as title helps to separate the posts, but of course it is not so clear now that we are dealing with a link. I merged it and we can see what others think.

PR #10, opened 2015-06-10
WEBSITE-282 Syntax highlighting for existing posts

gunnarmorling:
This adds syntax highlighting for existing posts via an Awestruct transformer that replaces `<pre>` blocks in the form of `<pre class="wikiPreformatted brush: java">...</pre>` with equivalent `<pre>`s highlighted using Coderay. Coderay is also what we use via Asciidoc, so the look is the same. Any thoughts?

### DIFF
--- a/_config/site.yml
+++ b/_config/site.yml
@@ -12,6 +12,8 @@ html_minifier: disabled
 # Disqus settings
 disqus_shortname: otnoitalerni-6
 
+legacy_post_syntax_highlighting: enabled
+
 ####################################################################################################
 # File merger
 ####################################################################################################
@@ -96,6 +98,7 @@ profiles:
     bootstrap_css_url: http://static.jboss.org/theme/css/bootstrap-community/2.3.1.3/bootstrap-community
     bootstrap_js_url: /javascripts/bootstrap-community
     #bootstrap_js_url: http://static.jboss.org/theme/js/libs/bootstrap-community/2.3.1.3/bootstrap-community
+    legacy_post_syntax_highlighting: disabled
   development:
     jborg_fonts_url: http://static.jboss.org/theme/fonts
     jborg_images_url: http://static.jboss.org/theme/images
--- a/_ext/legacy_post_code_highlighter.rb
+++ b/_ext/legacy_post_code_highlighter.rb
@@ -1,0 +1,47 @@
+##
+#
+# An extension of type transformer which modifies posts using the legacy syntax highlighting by replacing affected
+# <pre> blocks with an equivalent <pre> highlighted by CodeRay.
+#
+##
+module InRelationTo
+  module Extensions
+    class LegacyPostCodeHighlightingTransformer
+
+      def transform(site, page, input)
+
+        ext = File.extname(page.output_path)
+        if !ext.empty?
+          ext_txt = ext[1..-1]
+          if ext_txt == "html"
+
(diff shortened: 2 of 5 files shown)


hferentschik:
I like the basic idea, just wondering why you are not doing this as part of the erb file creation. We would just need to do this once, instead of on very single build. Since you are already using ruby it should be just a question of adding what `legacy_post_code_highlighter.rb` does to `importer.rb`. Or even better to `BlogEntry.rb`. The transformation could for example happen when the setter for `content` is called. Or just as part of `to_erb`.

gunnarmorling:
Ah, yes. Wasn't aware of that possibility, but it sounds very reasonable. I
will give it a shot.

hferentschik:
> Ah, yes. Wasn't aware of that possibility, but it sounds very reasonable. I will give it a shot.

Should be easy. This way we do the transformation once off. The README contains information about how to run the importer. I think in your case something like (from w/i in importer directory) should do:

    bundle exec ./importer.rb -s posts.pstore -o ../posts -ni -na

If you don't want to override the existing posts just chose a different output directory. If you want to just work with a subset of the posts or one specific one have a look at the `--lace` and `--limit` options. 

The latest pstore file is on the shared Dropbox folder.

emmanuelbernard:
the small advantage of this option is that _when_ we move to the next highlighter du jour, the code to update it is straightforward and in the transformer. I kind of like this exit strategy but maybe it can be done from the generated pre that Coderay uses?

hferentschik:
>  I kind of like this exit strategy but maybe it can be done from the generated pre that Coderay uses?

I think that should be possible. Or one just re-runs the generation. As it, it is taking up build time (if only marginal) which I would like to avoid.

gunnarmorling:
The HTML created by CodeRay looks like below. So we would have to remove all the spans upon a later update. Should be doable. Or one re-imports all the old posts if possible.

But then doing it in the transformer costs 3 sec. (1:34 min build time for a clean build on my machine vs. 1:31 on master). I think I'd slightly prefer the clearer mark-up in the ERB files and the easier updateability later on.

    <pre class="CodeRay highlight"><code data-lang="Java"><span class="annotation">@Entity</span>
    <span class="directive">public</span> <span class="type">class</span> <span class="class">Order</span> {

    <span class="annotation">@Id</span>
    <span class="annotation">@GeneratedValue</span>
    <span class="directive">private</span> <span class="type">long</span> id;

    <span class="annotation">@ElementCollection</span>
    <span class="directive">private</span> <span class="predefined-type">List</span>&lt;OrderLine&gt; orderLines = <span class="keyword">new</span> <span class="predefined-type">ArrayList</span>&lt;&gt;();

    <span class="comment">// getter, setters etc. ...</span>
    }

    <span class="annotation">@Embeddable</span>
    <span class="directive">public</span> <span class="type">class</span> <span class="class">OrderLine</span> {

    <span class="directive">private</span> <span class="predefined-type">String</span> item;
    <span class="directive">private</span> <span class="predefined-type">BigDecimal</span> price;

    <span class="comment">// getter, setters etc. ...</span>
    }</code></pre>

gunnarmorling:
Ah, btw. we also could have a property at the environment level, enabling the formatting only for production. Then you wouldn't even pay any penalty during development.

gunnarmorling:
Rebased and pushed one more commit disable highlighting in the "editor" profile.

PR #11, opened 2015-06-11
WEBSITE-299 Cloud cleanup

DavideD:
https://hibernate.atlassian.net/browse/WEBSITE-299

I've added a function in the normalize_tags.rb file.
This fucntion is going to be called during the import and allows to iterate over the tags of a post and return a different tag list.

At the moment, it is quite conservative, I've just reduced the dimensions of some of them like `Context and Dependency Injection` (`CDI`).

I've changed `Core Release` to `Release` and all the `Seam *` to `Seam`.
Scrolling through the erb file changes should give you an idea

I think we could also split some keywords in two, like:
```
JBoss AS 7, JBoss AS 6 => AS7, AS6
Hibernate Validator => Hibernate, Validator
Hibernate ORM => Hibernate, ORM
Hibernate OGM => Hibernate, OGM
Core Release => Hibernate, ORM, Release
JBoss Tools => JBoss, Tools
JBoss Asylum => JBoss, Asylum
...
```

I'm also tempted to remove spaces inside the tags:
```
Unified EL => UnifiedEL
Bean Validation => BeanValidation
```

Let me know what you think.


emmanuelbernard:
A few general comments

I don't think I mind space and uppercase letter in tags as long as they are rendered properly when used in URLs.
It is useful to be able to list all the tags when writing a blog so that one can reuses one rather than invent a new one. The drawback of only displaying the major tags is that they are not all visible ;) Not sure how to address that, create a special `/list-of-tag/` page? Or show the full list in editor mode?

That's more a philosophical debate but we should converge to a list of well known tasks we want.

Let me start with a few rules:

- one for tagging a release
- one for each product (Hibernate Search, OGM, ORM, Validator, Rich Faces etc)
- one for each spec (kind of a product). Not that Web Beans is the old name of CDI, maybe move
- one for podcasts (JBoss Asylum)
- one for more philosophical discussions ?


Final thought, the tags should probably be overall smaller on the page.

DavideD:
Note that If we agree before end about the keywords I can add some kind of validation for it.

> reate a special /list-of-tag/ page

There is already one: `/tags`
@emmanuelbernard if you could take a look at that list and let me know which keywords you want to keep or change that would be great.

> One for each product (Hibernate Search, OGM, ORM, Validator, Rich Faces etc)

I find `Hibernate ORM` a bit longer and redundant since the word Hibernate would appear everywhere, but I'm ok to keep it as a single tag.

This is the kind of tags I would like to have:

* `Hibernate`.
* `ORM`
* `OGM`
* `Search`
* `Validator`
* `Tools`
* `Shards`

* `JBoss`
* `Infinispan`
* `WildFly`
* `EAP`
* `JBoss AS`
* `Rich Faces`
* `Seam`

* `JBoss Asylum`
* `CDI`
* `Keynote`
* `JavaEE`
* `JSF`
...

> one for tagging a release

I would go with `Release` :)
And then, if it's easy, I would create some rest URLs `/{project}/Release/{Version}` and `/{project}/release`

> one for each product (Hibernate Search, OGM, ORM, Validator, Rich Faces etc)

> one for podcasts (JBoss Asylum)
 
Sounds Good

> one for more philosophical discussions ?

Maybe `Discussion` or `Proposal`

Sanne:
> And then, if it's easy, I would create some rest URLs /{project}/Release/{Version} and /{project}/release

I like the idea but these should be implemented as either:
 - have an additional element in the <head> section of the page
 - a redirect

Let's do it after the mvp release?

DavideD:
> I like the idea but these should be implemented as either:
> 
>     have an additional element in the section of the page
>     a redirect
> 
> Let's do it after the mvp release?

Yes, this is a separate issue anyway

emmanuelbernard:
BTW, let's keep the original tags in the erb file metadata by creating an original_tags property alongside the tags one.

DavideD:
> BTW, let's keep the original tags in the erb file metadata by creating an original_tags property alongside the tags one.

OK

DavideD:
I've applied the following changes

* Reduce the font size of the tags cloud
* Split words starting with 'Hibernate ' or 'JBoss ', example: Hibernate OGM => Hibernate, OGM
* Add original_tags
* Update README
* Added a .gitignore

Sanne:
I'll push it to staging ..

emmanuelbernard:
So here is my list with small remaining questions.

I like `Hibernate Search` as opposed to `Hibernate` `Search` if nothing because the title In relation to Hibernate Search is more accurate. So someone really willing to split words will have to come with arguments. For the same reason, I like the proper upper cases and spaces tags.

This list means any variation (upper case, space etc). I hopefully listed them all without typo but that's unclear.

I think one or two go from one tag to several. It's possible that this conversion, returns the same tag several time for a given blogpost, it needs to be deduplicated.

* `Hibernate`: `Hbernate` => remove except in case below for ORM.
* `Hibernate ORM`: `Core Release`, `Criteria Queries`, `Databases`, `PostgreSQL`, `MySQL`, `Hibernate Metamodel Generator`, `ORM`, `orm43`
** if `Hibernate` does not have another product tag, we should move it to ORM otherwise ignore
* `Hibernate OGM`: Hibernate OGM 101, OGM
* `Hibernate Search`
* `Hibernate Validator`:  hv51,
* `Hibernate Tools`
* `Hibernate Shards`: `Shards`

* `Release`: `Core Release`, `ReleaseNotes`

* `JBoss Tools`: `JBoss Tools Eclipse`
* `Infinispan`
* `WildFly`
* `JBoss EAP` (from `EAP`?): @Davide I did not see this one, was it invented?
* `JBoss AS` should we merge it with WildFly? Probably no:  JBoss AS, JBoss AS 6, JBoss Modules, JBoss AS 7, `JPA Hibernate JBoss AS`, `as7`,`jbossas7`, `as7_openshift`, `jboss5`
* `RichFaces`: `Rich Faces 2`, `RichFaces 4`
* `Seam`: `Seam News`, `Seam 3`, `Seam 3 News`, `Seam Forge`, `Seam Wiki`, 
* `AeroGear`
* `Arquillian`
* `Ceylon`: `Ceylon IDE M1`, `Introduction to Ceylon`
* `PicketLink`
* `Weld`: `Weld Extensions`, 

* `Event`: Events, `Conference`, `conference`, `EclipseCon`, `openblend`, jbug, jbw2011keynote,  jbwkeynote2011, jug, Keynote
* `JBoss Asylum`: `Asylum`, JBossAsylum, 

* `JPA`: `JPA 2`, `JPA Hibernate JBoss AS`, jpa21, 
* `CDI`: 'Contexts and Dependency Injection`, `Web Beans`, `Web Beans Sneak Peek`
* `Bean Validation`: `Beam Validation`, `Bean Validation Sneak Peek`
* `Java EE`: `EE`, `EE6`, `EE6 Wishlist`, `ICEfaces ICEfaces2`, Java EE 6, JavaEE, 

* `Discussion`: `Design`, `Git`, `Gradle`, `Maven`, `Persistence`, `Proposal`, `Review`, `jDocBook`
* `Off topic` (or Misc?): `Photography`

Should we keep the following or not?

* `JBoss Forge`: Forge
* `Eclipse` (probably worth keeping?)
* Should we remove News (very generic and not useful I think)

* No tag: Ajax, Announcement, AuthorDoclet, Award, BLOB, Book, Books, Code Coverage, Comet, Community, ConnectionProvider, Cordova, DeltaSpike, `Dependency Management`, `Development`, `Devoxx`, Dukes Choice Award, EJB, EclEmma, Embedded JBoss, Errai, EventListeners, ExtendedDataTable, Facelets, Geeky Java Stuff, Granite DS, HTML5, Hints, Interceptors, Interoperability, JBoss, JBoss Portlet Bridge, JDO, JMS, JSF, JSF 2, JUDCon, Java Server Faces, JavaOne,  JavaServer Faces, Jigsaw, Jokre, Linux, London, Lucene,  Maintenance, Marshalling, Media, Metawidget, Mobile, Modules, Mojara 2, OpenJDK, OpenShift, POH5, Paris JUG, PartialViewContext, Payasos, Performance, Portable Extensions, Push, REST, RESTEasy, SeamGen, Security, `Solder`, `Spring`, `tattletale`, `Testing`, TorqueBox, `Unified EL`, `Versioning`, `Web Frameworks`, `WebSockets`, Wicket, `XML Hell`, calendar, dataTable, `deployers`, `vfs`, microcontainer, `dirtiness`, `feature`, java, jbosscentral, jdf, 

A lot of the removed ones are either:

* they don't add value or mainly duplicate of non deleted ones
* they were referencing one or two posts

Final request for comment.

Sanne:
Stuff like 'jbw2011keynote' was a big cross-blog initiative with various cross-links and articles to explicitly suggest to search for that tag.

I realize that this specific keyword might not be anymore relevant.. but what's the intention to deal with say a similar event cycle in the near future?
Also, there is a sad loss of historical information.

What I'm missing here is what is the mission. Do we want to stick to a small set of keywords and then enforce that? What is the goal?

DavideD:
>  like Hibernate Search as opposed to Hibernate Search if nothing because the title In relation to Hibernate Search is more accurate. So someone really willing to split words will have to come with arguments. For the same reason, I like the proper upper cases and spaces tags.

Ok, I just proposed the split to have shorter word but it looks nicer with the long words

> Should we remove News (very generic and not useful I think)

Isn't used currently by other sites to keep track of new posts?

Sanne:
> Should we remove News (very generic and not useful I think)

>Isn't used currently by other sites to keep track of new posts?

Yes, and therefore I add it all the time ;-)

Sanne:
staging.in.relation.to starts to have a nice shape :+1:  I like the current tag cloud already.

posts/Emmanuel/2011-08-26-HibernateOGMAndRedHatAtTheOpenWorldForumAndOSDC.html.erb
```diff
@@ -3,7 +3,8 @@ title: "Hibernate OGM and Red Hat at the OpenWorldForum and OSDC"
 author: "Emmanuel Bernard"
 blogger_name: "Emmanuel"
 creation_date: "26-08-2011"
-tags: [Hibernate OGM,conference]
+original_tags: [Hibernate OGM,conference]
+tags: [conference,Hibernate,OGM]
```
gunnarmorling: "OGM" should be "Hibernate OGM" I guess?

gunnarmorling:
> I like "Hibernate Search" as opposed to "Hibernate" "Search"

Yes, absolutely.

gunnarmorling:
@emmanuelbernard I think your list looks good. What is the idea in general: Leave the original tags on the posts and only use that aggregated tag list for the "tag cloud"? Or will the original tags not be visible on the new page any longer at all (only kept for reference in a never-exposed field in ERB files.

I will also need to process that aggregation function as part of the redirection rules creation.

gunnarmorling:
> Hibernate OGM: Hibernate OGM 101, OGM

I had used "Hibernate OGM 101" as tag for all the tutorial parts (as opposed to release announcements). But they link to all the previous instalments, so it should not be a loss.

gunnarmorling:
> Discussion: Design, Git, Gradle, Maven, Persistence, Proposal, Review, jDocBook

Maybe "Build" for "Gradle" and "Maven"? The others look a bit more "philosophical".

gunnarmorling:
> No tag: [...]

Is this to say that these tags should be removed? If so, I would at least move/merge the following:

Dependency Management, SeamGen, Solder -> Seam
DeltaSpike, Interceptors, Portable Extensions -> CDI
JavaOne, Devoxx, JUDCon, Paris JUG, London -> Conference
JSF, JSF 2, Java Server Faces, PartialViewContext, Mojara 2 -> JSF

Tbh. I am having doubts about removing tags altogether (if that's the plan, it's not clear to me). +1 for consolidation and clean-up, but I would dislike if e.g. the "Web Frameworks" posts would have no tag at all afterwards. If size of the tag cloud is the driving issue, it could be limited to the "n most-often used tags".

Sanne:
> Tbh. I am having doubts about removing tags altogether (if that's the plan, it's not clear to me). +1 for consolidation and clean-up, but I would dislike if e.g. the "Web Frameworks" posts would have no tag at all afterwards. If size of the tag cloud is the driving issue, it could be limited to the "n most-often used tags".

+1 a tag cloud to be effective needs to be automatically computed on a large variety of "freely defined" keywords. Relative Frequency is one of the scale factors, but age should be included in the formula.
I think that would work better than manual trimming, and is more resilient than in-house rules and conventions for the future.

emmanuelbernard:
> Stuff like 'jbw2011keynote' was a big cross-blog initiative with various cross-links and articles to explicitly suggest to search for that tag.
> 
> I realize that this specific keyword might not be anymore relevant.. but what's the intention to deal with say a similar event cycle in the near future?
> Also, there is a sad loss of historical information.

I find the alternative to link backward and forward between different posts of a given series acceptable for such situation.

emmanuelbernard:
> What I'm missing here is what is the mission. Do we want to stick to a small set of keywords and then enforce that? What is the goal?

Today we have a tag cloud that is at best 1/3 of the total number of tags we have. It's impractical to show them all. The side effect is that people create new related tags but with different spellings. Which means that the aggregation of related information does not really work.

So this aggregation work is here to bring order to this chaos.

The second point is to figure out why tags are useful and why someone would click on it. My theory is that tags are here to categorize related content under a banner. What banner do we really need? I argued that project names, specs and a few others is what really matters in term of usability.

emmanuelbernard:
>>> Should we remove News (very generic and not useful I think)

>> Isn't used currently by other sites to keep track of new posts?

> Yes, and therefore I add it all the time ;-)

I would like to challenge that. It *used* to be useful at some point in the history of that aggregation logic. But I don't think it is even used anymore. From what I have seen, planet.jboss.org simply aggregate all posts from its blogs and that's it.
If you can find a fresh info contradicting what I see, we can reconsider that dropping.

But to be fair, What blog entry do we *not* want included in the aggregation engine? Practically none. In which case we are better off adding a tag synthetically to please the aggregation non sense remove that meaningless `news` tag.

emmanuelbernard:
> @emmanuelbernard I think your list looks good. What is the idea in general: Leave the original tags on the posts and only use that aggregated tag list for the "tag cloud"? Or will the original tags not be visible on the new page any longer at all (only kept for reference in a never-exposed field in ERB files.

My idea was to not display the old tags at all to be honest. What would be the benefit of exposing a non clickable tag? I don't think anyone decides to read a blog based on the tags.

> I will also need to process that aggregation function as part of the redirection rules creation.

yes that would be preferable

emmanuelbernard:
> > Discussion: Design, Git, Gradle, Maven, Persistence, Proposal, Review, jDocBook

> Maybe "Build" for "Gradle" and "Maven"? The others look a bit more "philosophical".

We could, but we would end up with non core tags popping up. Should they really be separated? If we do `Build`, should we do `tools`, `processes`, etc etc?
I saw `Discussion` as kind of a generic container.

Sanne:
>  From what I have seen, planet.jboss.org simply aggregate all posts from its blogs and that's it.
If you can find a fresh info contradicting what I see, we can reconsider that dropping.

I don't have contradicting information. Since I always use that tag, I'm not sure if that's the cause or the effect.
But I agree we always want it to appear, so we don't need one to be explicitly defined by the user, still maybe the feed needs it for planet to show it.

emmanuelbernard:
>> No tag: [...]

> Is this to say that these tags should be removed? If so, I would at least move/merge the following:
> 
> Dependency Management, SeamGen, Solder -> Seam

These blog posts are already tagged `Seam` so removing them would be a noop. But if you prefer we can include them to "document" the migration better.

> DeltaSpike, Interceptors, Portable Extensions -> CDI

These blog posts are already tagged `CDI` (by ricochet) so removing them would be a noop. But if you prefer we can include them to "document" the migration better.

> JavaOne, Devoxx, JUDCon, Paris JUG, London -> Conference

These are already marked `Event` which encompasses conferences and JUGs and vJUGS etc. Do we need a separate `Conference` tag ora rather a better name than `Event`. `Event` is what captured things best when I thought about the problem.

> JSF, JSF 2, Java Server Faces, PartialViewContext, Mojara 2 -> JSF

OK, I've updated my initial post.

> 
> Tbh. I am having doubts about removing tags altogether (if that's the plan, it's not clear to me). +1 for consolidation and clean-up, but I would dislike if e.g. the "Web Frameworks" posts would have no tag at all afterwards. If size of the tag cloud is the driving issue, it could be limited to the "n most-often used tags".

We already do the limitation and a lot of tags are dropped form view (see discussion above).

emmanuelbernard:
> But I agree we always want it to appear, so we don't need one to be explicitly defined by the user, still maybe the feed needs it for planet to show it.

If our migration breaks planet, we will do that hidden tag trick.

DavideD:
The cloud looks much better now and it displays all the tags
There are plenty of posts without tags though.

You can see the list of tags at `localhost:4242/tags`

emmanuelbernard:
@DavideD Davide I tried to apply your change but would not rebase cleanly. I starts to get out of my comfort zone, can you rebase?

hferentschik:
> So this aggregation work is here to bring order to this chaos.

+1

> The second point is to figure out why tags are useful and why someone would click on it. My theory is that tags are here to categorize related content under a banner. What banner do we really need? I argued that project names, specs and a few others is what really matters in term of usability.

+1

BTW, I also see no problem in some tags not having a tag. A tag is not mandatory. I don't have to invent obscure tags just to add a tag to a blog which might fall a bit out of the usual content pattern.

Could we not also make the site searchable by adding a Google Search Box which searches just our site?

DavideD:
I've rebased everything but the tags don't appear anymore.

Does it only happen to me?

gunnarmorling:
Same for me, author list and tag cloud are gone.

hferentschik:
>I've rebased everything but the tags don't appear anymore.

Which profile are you building? The default is the 'editor' profile now which skips tag and author splitting. It just processes the main posts. Do you use 'rake gen[development]' or 'rake preview[development]'?

hferentschik:
FYI, HEAD of master is pushed to http://staging.in.relation.to/ and looking good. That said, pushing to to the _staging_ branch, did not trigger the CI build. I had to start it manually. @Sanne !?

gunnarmorling:
> The default is the 'editor' profile now which skips tag and author splitting. 

Ah, yes. It's working with the "development" profile. Thanks.

DavideD:
Thanks, it works for me as well.

Maybe now the cloud it's too small. I think I prefer the original CSS.
I'll let the commit (https://github.com/DavideD/in.relation.to/commit/8d0cdb7cde9ea42675a8e1805595d5231d69ca52) in the pull request so that you can check the difference.

hferentschik:
Ok, I re-based once more and pushed it to http://staging.in.relation.to/. Looking good to me. I think we should merge it. We can always adjust later. For example, I think the _Release_ tag should actually be _Releases_. That makes more sense when looking at the tag cloud.

hferentschik:
ok, merged. We can improve from here.

PR #12, opened 2015-06-11
WEBSITE-287 Asciidoc / Markdown examples

emmanuelbernard:
Put as "Mister Test" under the Test subdirectory.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,24 @@
+## Bundler
+_bin
+.bundle
+
+## Awestruct
+_tmp
+_site
+.awestruct
+.sass-cache
+
+## OS related
+*.DS_Store
+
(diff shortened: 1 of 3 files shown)


hferentschik:
Cool, so it works out of the box :-) I was expecting that there is still some pixie dust missing. Given that it works, I am wondering whether this needs to be put into master? In the end it would need to be removed anyways. Is there a value to have them there for the dev process?

emmanuelbernard:
Put it in, it's useful for tests, I need to put them on all PR I'm doing. basically

hferentschik:
Merged, just made a change to the markdown example which was using `==` instead of `##`. Either way, both formats seem to work :-)

PR #13, opened 2015-06-11
WEBSITE-292 Atom feed generation

hferentschik:
Some notes: 

* I actually removed the checked in `atomizer.rb` and its template. I rely on the ones provided by Awestruct. The extension itself should be usable out of the box a remaining issue is the template. This we might need depending on the points below
* There were two things which I thought were "wrong" with the initial feed
  * The blog url which ended in `.html`. This is actually fixed by the change to the Posts extensions with the additional benefit that we don't need the `Indexifier` in this case
  * When looking at the feed in Firefox, I noticed initially that the displayed dates are identical. I thought this was a bug, but I don't think so anymore. Basically Atom feeds define a publish date and a last change date. The publish date comes form the blog entry meta data, whereas the last changed date is determined by the last modified time of the actual file. Given that we generate all erb files in one go, the last modified time will be the same initially. Later down the track though, it might make sense to have this distinction. If a blog entry gets updated it would get pushed into the feed again. Hope this make sense. If we just want to rely on the published date, we need to add back the atomizer template and make a small change there. Basically open for discussion
* Last but not least, I choose `in-relation-to.atom` as a more descriptive name of the atom feed. Again, up for discussion.


emmanuelbernard:
On the name of the feed. I would prefer to remain domain name agnostic. Even now me might decide to redirect in.relation.to to blog.hibernate.org (better Google ranking). For sure it would be nice to keep the options for later. So a generic blog.feed name would be nice. It also can be declined `/author/blog,feed` etc.

emmanuelbernard:
On the atom update field, that's a interesting question. Back in the day I ended up syncing creation time and update time because it was creating headaches with the feed readers (like everything marked as unread again from the origin of time). I'm fairly certain readers don't fully respect the update field basically.

The other problem is how is this update time computed? What if I git clone, does it change?

hferentschik:
> I'm fairly certain readers don't fully respect the update field basically.

Ok, not sure how readers handle that. Have not used a feed for a long time.

> The other problem is how is this update time computed? What if I git clone, does it change?

That might indeed be an issue after a quick check on how got behaves at cloning. I guess i sync creation and update time in this case. Need to rebase anyways as well.

hferentschik:
> What about all the other pages generated like pagination, tags, authors? We would need to change their extensions too no? Or leave the Indexifier on.

I think it works, but I double check. In the worst case we leave the Indexifier enabled. For the main posts it will be a no-op.

hferentschik:
> On the name of the feed. I would prefer to remain domain name agnostic. 

Interesting. I personally prefer the clear name. See also below

> Even now me might decide to redirect in.relation.to to blog.hibernate.org (better Google ranking)

Not sure about this. Not sure how far we have to go with Google ranking considerations. I like `in.relation.to`, because it is a twist. `blog.hibernate.org` is just plain boring.

emmanuelbernard:
@hferentschik There is something I don't grok, why do you thing `in-relation-to.atom` is clearer than `blog.atom`?

hferentschik:
@emmanuelbernard, I pushed an update. I went for `blog.atom` as feed name and use a modified template which just uses the _updated_ node (aligning with ceylon-lang blog).

Regarding _Indexifier_ it is not needed in this case. The paginated pages use index pages as well.

PR #14, opened 2015-06-11
WEBSITE-263 Introduce editor profile

emmanuelbernard:
It removes 25s to the full generation (1/3 of the time) and more importantly, it avoid the issue described in https://hibernate.atlassian.net/browse/WEBSITE-318.

@gunnarmorling can use the editor state in pipeline to remove the Coderay generation at editor time.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,24 @@
+## Bundler
+_bin
+.bundle
+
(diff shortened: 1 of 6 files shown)


hferentschik:
Merged with some minor improvements and update to README.

PR #15, opened 2015-06-12
WEBSITE-297 Attachments

DavideD:
https://hibernate.atlassian.net/browse/WEBSITE-297

Render a list of attachments at the bottom of the blog entry.


hferentschik:
Looking basically good, but I think you did skip the image processing in the generated files, so that image links are not adjusted. I will re-run the erb creation and amend the last commit ...

DavideD:
> I think you did skip the image processing in the generated files,

Not sure I'm following...wha do you mean?

>  I will re-run the erb creation and amend the last commit ...

No need to remove the last commit, I've just realized that this solution won't work when multiple posts with an attachment appears in the same page. That's because they will have the same anchor name.

hferentschik:
> Not sure I'm following...wha do you mean?

No problem, it was basically an unrelated issue. I fixed it and will create a new issue. 

> No need to remove the last commit, I've just realized that this solution won't work when multiple posts with an attachment appears in the same page. That's because they will have the same anchor name.

> No need to remove the last commit, I've just realized that this solution won't work when multiple posts with an attachment appears in the same page. That's because they will have the same anchor name.

Ahh, do you mean on the author/tag lists? Will you create an update to this pull request then? I'll commit the changes to the bug I found separately.

DavideD:
At the moment, if you show all the posts (in the home for example), the attachments appear at the bottom of the post.

hferentschik:
Btw, maybe we don't even need that. If the links to the attachments work, the most important thing is working. The issue is also more of a question.

DavideD:
I've added two commits that will make the anchor name unique using the slug as prefix

hferentschik:
Re-based and merged.

PR #20, opened 2015-06-17
Updating Gemfile.lock to Bundler 1.10.4

gunnarmorling:


hferentschik:
Merged. As expected you just get a warning, which is actually a good thing. I also updated the docs to mention the requirement of Bundler version >= 1.10

emmanuelbernard:
I don't understand why that is useful, can you explain it to me. bundler is needed to understand Gemfile, so why have that? And do we have to enforce 1.10.4, can't we allow >= 1.10?

gunnarmorling:
Bundler adds this entry as of >= 1.10 (which we require as minimum now).
Why, I don't know :)

Older 1.10.x versions will leave it untouched, but newer versions will
update it. Hence it makes sense to have the latest version in there,
otherwise those of us actually using the latest bundler will always see a
change to the file in their working area. Update to 1.10.x is useful as
older Bundler versions will remove the entry, again giving you an unwanted
change in your working area.

All in all an exercise in pleasing some tool to get peace of mind :/

hferentschik:
> All in all an exercise in pleasing some tool to get peace of mind :/

+1 That's pretty much it. According to the Bundler guys they will need this information for some feature changes down the road. I kind of get the value in it, but it for sure creates a fair bit of uproar. As Gunnar is saying, best to set it to the latest possible version, since a version > 1.10 but < 1.10.4 will just print a warning, but proceed. Problematic are pre 1.10 versions since they will remove the metadata gain leaving you again with a changed file in your repo.

PR #24, opened 2015-06-18
WEBSITE-336, WEBSITE-337

gunnarmorling:
Importing author bios and showing their gravatar (I found most, but some are missing; The active authors are there basically, only @DavideD is missing).

I don't show any of website, occupation or location atm. It's imported as meta-data though. We could show links to website, blog and twitter. Let me know if you'd like any of that.


emmanuelbernard:
Applied. Though I don't like how it shows under phone sized screens. Should we not show bios at all in the phone version? @DavideD can help. Leaving the issue open until we resolve that.

PR #26, opened 2015-06-18
WEBSITE-291 Remove unnecessary files, add doc on how to blog

emmanuelbernard:


README.adoc
```diff
@@ -96,6 +96,119 @@ you should use:
...
+For example `2015-06-30-multitenancy-and-current-session.adoc`.
+
+Here is a sample blog template you can start with:
+
+[source]
+.An Asciidoctor blog post
+....
+= Let me tell you about...
+:author: Emmanuel Bernard
+:awestruct-tags: ["Blog", "Test"]
```
gunnarmorling: Why not just "tags" and "layout"?

README.adoc
```diff
@@ -96,6 +96,119 @@ you should use:
...
+----
+
+== More info
+
+You can read more on the Asciidoctor syntax at http://asciidoctor.org.
+
+Happy blogging.
+....
+
+=== Metadata
```
gunnarmorling: The duplication of example post and actual content blows it up quite a bit.

gunnarmorling:
Looks good apart from my two remarks above.

README.adoc
emmanuelbernard: that's how the Asciidoctor/Awestruct integration is done. Only some tags and auto wired. Otherwise you need to prefix them with `awestruct-`

README.adoc
emmanuelbernard: blows up?

Anyways, yes I would like to have something a bit better here but the rendering is better when that's not a code example. And if someone copy paste the code example, he's got all the rules.

Ideas welcome.

emmanuelbernard:
Looks like I applied it by accident. We can keep talking to improve it of course.

PR #27, opened 2015-06-19
WEBSITE-320, WEBSITE-334, WEBSITE-345: Displaying authors and tags consistently in the side bare and below post titles

gunnarmorling:


hferentschik:
+1 for most of the changes. The site feels more consistent now. I am not so sure, whether I like the switch of tags with authors. I'll merge it anyways, so everyone can have a look.

hferentschik:
merged.

PR #28, opened 2015-06-19
WEBSITE-346 Cleanup of site profiles

hferentschik:
Some notes. Most of the cleanup makes imo sense regardless. The point worth discussing the removing of the base_url. This makes staging and production identical and hence the staging profile could be removed. The only problem with the site in this case is the atom feed generation which needs an absolute URL. Since we control the template for the atom feed, I just went ahead and hard coded in.relation.to in there. Thoughts?

BTW, to test this against staging, one needs to change the job config to build 'production', since there is no dedicated staging profile anymore.

### DIFF
--- a/_config/site.yml
+++ b/_config/site.yml
@@ -90,64 +90,40 @@ asciidoctor:
 ####################################################################################################
 # Update resource URLs based on the profile
 profiles:
-  editor:
-    jborg_fonts_url: http://static.jboss.org/theme/fonts
-    jborg_images_url: http://static.jboss.org/theme/images
-    jborg_js_url: http://static.jboss.org/theme/js
-    project_images_url: /cache/static.jboss.org/images/hibernate
-    bootstrap_css_url: http://static.jboss.org/theme/css/bootstrap-community/2.3.1.3/bootstrap-community
-    bootstrap_js_url: /javascripts/bootstrap-community
-    #bootstrap_js_url: http://static.jboss.org/theme/js/libs/bootstrap-community/2.3.1.3/bootstrap-community
-    legacy_post_syntax_highlighting: disabled
-  development:
-    jborg_fonts_url: http://static.jboss.org/theme/fonts
-    jborg_images_url: http://static.jboss.org/theme/images
-    jborg_js_url: http://static.jboss.org/theme/js
-    project_images_url: /cache/static.jboss.org/images/hibernate
-    bootstrap_css_url: http://static.jboss.org/theme/css/bootstrap-community/2.3.1.3/bootstrap-community
-    bootstrap_js_url: /javascripts/bootstrap-community
-    #bootstrap_js_url: http://static.jboss.org/theme/js/libs/bootstrap-community/2.3.1.3/bootstrap-community
-  test:
-    jborg_fonts_url: http://static.jboss.org/theme/fonts
-    jborg_images_url: http://static.jboss.org/theme/images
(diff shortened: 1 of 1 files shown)


hferentschik:
On a tangent, we use the blog url as an id for a feed entry. This is according to this post - http://stackoverflow.com/questions/150741/how-do-you-generate-a-good-id-in-atom-documents - not a good strategy. Best strategy would probably be to have a GUID in the metadata of a blog entry, but how would one get it in there?

hferentschik:
Sorry, I was wrong. I tested against a system which was not properly deployed. I'll add back the base_url, but keep the rest of the cleanup

hferentschik:
Ok, now the pull request just does a profile cleanup, leaving the base_url

_config/site.yml
```diff
@@ -90,64 +90,41 @@ asciidoctor:
...
-    jborg_js_url: http://static.jboss.org/theme/js
+    << : *base
+    legacy_post_syntax_highlighting: enabled
     project_images_url: http://static.jboss.org/images/hibernate
-    bootstrap_css_url: http://static.jboss.org/theme/css/bootstrap-community/2.3.1.3/bootstrap-community
-    bootstrap_js_url: /javascripts/bootstrap-community
-    #bootstrap_js_url: http://static.jboss.org/theme/js/libs/bootstrap-community/2.3.1.3/bootstrap-community
     wget:
-      #we don't need to download cached resources when in production
+      #we don't need to download cached resources when on staging
```
gunnarmorling: That's production?

_config/site.yml
```diff
@@ -90,64 +90,41 @@ asciidoctor:
...
     jborg_images_url: http://static.jboss.org/theme/images
     jborg_js_url: http://static.jboss.org/theme/js
-    project_images_url: http://static.jboss.org/images/hibernate
+    project_images_url: /cache/static.jboss.org/images/hibernate
     bootstrap_css_url: http://static.jboss.org/theme/css/bootstrap-community/2.3.1.3/bootstrap-community
     bootstrap_js_url: /javascripts/bootstrap-community
-    #bootstrap_js_url: http://static.jboss.org/theme/js/libs/bootstrap-community/2.3.1.3/bootstrap-community
+
+
+  development:
```
gunnarmorling: Is that still needed at all then?

emmanuelbernard:
Don't bother with the id. A permalink is good enough. we don't care much in practice.

hferentschik:
> Don't bother with the id. A permalink is good enough. we don't care much in practice. 

ok

_config/site.yml
hferentschik: Technically not in _site.yml_. We make still use of the development profile, but that is happening in the pipeline configuration. Right now there is no difference in the two profiles as defined in _site.yml_.

I think in a perfect world we would make sure that all these site specific properties would be in _site.yml_. For example there would be a property which enable/disable and potentially specifies the age limit of posts to process and one property which would determine whether or not to process tags/authors. This way there would be one place to look for these things. Right now, you really need to look at various places and piece the things together. 

For now I think, I'll just remove the development profile from _site.yml_.

hferentschik: > For now I think, I'll just remove the development profile from site.yml.

Ahh, that I can of course not do, since even though identical to  the editor settings, they still need to be defined.

hferentschik:
merged

PR #35, opened 2015-06-22
WEBSITE-314 Importing in.relation.to into PStore file

hferentschik:
- Adding BZIP2 version of site as posts.pstore.bz2
- Updating importer readme
- Updating test disqus id to otnoitalerni-8
- Disabling spam detection in wxr_exporter, since site is pretty much spam free to avoid false positives
- Adding new ERB files
- Updating Apache redirect config


hferentschik:
This change can already be viewed on staging (there seems to be an issue w/ comments. I am looking into it)

posts/Davide/2015-06-02-hibernate-ogm-42-final-is-out.html.erb
```diff
@@ -0,0 +1,83 @@
...
+lace: http://in.relation.to/42822.lace
+
+layout: blog-post
+
+disqus_thread_id: http://in.relation.to/2015/06/02/hibernate-ogm-42-final-is-out
+---
+<div id="documentDisplay" class="documentDisplay">
+
+
+<p class="wikiPara">
```
gunnarmorling: Is there a reason for keeping the legacy CSS classes or should we enhance the importer remove them? We don't style them anyways, so they only increase the size of the delivered HTML. Thoughts? Can be done separately of course.

gunnarmorling:
Rebased and applied. Note that I prepended a commit and reworked yours in order to have a stable order of entries in .htaccess_posts. That way the changes to this file in your commit are just limited to the redirects for the added posts.

hferentschik:
> Rebased and applied.

Thanks

> Note that I prepended a commit and reworked yours in order to have a stable order of entries in .htaccess_posts. 

Nice one. I noticed the redirects diff was quite large. Sorting the entries seems to be a good idea.

hferentschik:
> Is there a reason for keeping the legacy CSS classes 

Maybe. I was thinking that at some stage we might want or have to adjust styles depending on whether we are dealing with old posts or new ones. Maybe not, but for now I think it does not hurt to keep them. Also, as you say, it can be done later on.

PR #39, opened 2015-06-23
WEBSITE-340 Setting the right page.title to make sure the HTML header…

hferentschik:
… title can be properly set

- Updating the HAML template to avoid double "In relation to - In relation to"
- Adding additional option to Paginator to set the page title and making sure Splitter passes on the right title
- Removing unused SplitLinker


hferentschik:
thanks

PR #43, opened 2015-06-24
WEBSITE-354 Adapting redirect files to being used via httpd.conf inst…

gunnarmorling:
…ead of .htaccess which needs leading slashes


Sanne:
thanks, merged it.

PR #45, opened 2015-06-24
WEBSITE-362 Changing <h1>/<h2> within posts to <h3> for the most rece…

gunnarmorling:
…nt posts to make home page nicer

As discussed via mail automating this would be nice. Still did it manually for the posts on the first two pages to make the first impression of the new blog a bit nicer.


hferentschik:
merged

PR #47, opened 2015-06-25
WEBSITE-364 Giving some love to pagination bar

gunnarmorling:


gunnarmorling:
Hold on, something is wrong with the rendering of titles. Looking...

_config/site.yml
```diff
@@ -98,6 +98,7 @@ profiles:
     js_minifier: disabled
     html_minifier: disabled
     legacy_post_syntax_highlighting: disabled
+    ignore_older_than_days: 30
```
hferentschik: nice one. Do you know whether one can pass in (aka override) these values form the command line. In development mode I often want to see a bit more and would love to change this more dynamically.

_ext/link_renderer.rb
```diff
@@ -0,0 +1,59 @@
...
+# pagination extension to be enabled.
+#
+##
+module InRelationTo
+  module Extensions
+    class PaginationLinkRenderer
+
+      module BootstrapPaginationLinkRenderer
+
+        def bootstrap_links
```
hferentschik: Gee, where did you dig this one out? Did you write it from scratch?

gunnarmorling:
Ok, force pushed an update. Originally I had removed the copy paginator.rb, but it has some customization compared to the original one from Awestruct we apparently need. So with the update I am leaving paginator.rb in place, only adding the other extension for the Bootstrap-style links.

_ext/link_renderer.rb
gunnarmorling: I took the original links method as an example and adapted it to use the Bootstrap CSS classes.

_config/site.yml
gunnarmorling: I don't know whether that's possible (can't find anything in the docs, so probably not). In development it's set to one year, but of course you can alter it there as you like.

_ext/link_renderer.rb
hferentschik: sweet :-)

_config/site.yml
hferentschik: > but of course you can alter it there as you like.

Sure, just I need to remember not to commit it

hferentschik:
Awesome, loving the new L&F. The pagination links where high on my list of required UI improvements. Very nice pull request on many levels.

_config/site.yml
DavideD, line 101: By the way, would it be possible to change it so that it keeps the latest 30 posts?
I would like to be able to specify the number of posts instead of the date.

gunnarmorling, line 101: Should be possible. Feel free to change the option if you like. Otherwise just make sure you blog enough so there are plenty of posts in the last 30 days ;)

DavideD, line 101: Since it's already done, I would keep it like this.

We can create an issue for this in the future if the need arise.

PR #49, opened 2015-06-25
Redirects deployer, httpd configuration

Sanne:


gunnarmorling:
We should rename the folder from "redirects" to something like "server-config", reflecting the more general semantics it has now. Don't forget to adapt .awestruct_ignore.

redirects/deploy.sh
```diff
@@ -0,0 +1,21 @@
...
+scp staginginrelationto.conf ci.hibernate.org:~
+
+cat file-opening-production.conf > inrelationto.conf
+cat .htaccess_misc >> inrelationto.conf
+cat .htaccess_bloggers >> inrelationto.conf
+cat .htaccess_posts >> inrelationto.conf
+cat file-closing.conf >> inrelationto.conf
+scp inrelationto.conf ci.hibernate.org:~
+
+ssh -t ci.hibernate.org "sudo cp staginginrelationto.conf /etc/httpd/conf.d/staginginrelationto.conf && sudo cp inrelationto.conf /etc/httpd/conf.d/inrelationto.conf && sudo service httpd restart"
```
gunnarmorling: Better use reload than restart?

redirects/file-opening-production.conf
```diff
@@ -0,0 +1,17 @@
+<VirtualHost *:80>
+  ServerName in.relation.to
+  ServerAdmin webmaster@localhost
```
gunnarmorling: Or our actual admin address?

redirects/deploy.sh
Sanne: thanks for that. when did they change it to 'reload' ?! I was looking for the 'graceful' but it was not accepted. I guess thanks systemd..

redirects/file-opening-production.conf
Sanne: +1

Sanne:
I've applied the suggestions and merged it.. thanks!

hferentschik:
> We should rename the folder from "redirects" to something like "server-config", reflecting the more general semantics it has now

+1

PR #51, opened 2015-06-26
WEBSITE-289 Add disqus comments counter

DavideD:
https://hibernate.atlassian.net/browse/WEBSITE-289

I've already applied the changes in staging, if you guys want to have a look

### DIFF
--- a/_layouts/blog-post.html.haml
+++ b/_layouts/blog-post.html.haml
@@ -19,20 +19,4 @@ title_partial: blog/title-post.html.haml
   .span10
     .blogpost
       = partial( 'blog/post.html.haml', {"post" => page} )
-    #disqus_thread
-    %script{:src=>javascript, :type=>'text/javascript'}
-      var disqus_shortname = '#{site.disqus_shortname}';
-      var disqus_identifier = '#{page.creation_date}-#{page.slug}';
-      var disqus_title = '#{page.title}';
-      var disqus_url = '#{page.disqus_thread_id}';
-
-      (function() {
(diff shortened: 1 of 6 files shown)


DavideD:
It seems there are some problems with some posts without comments: http://staging.in.relation.to/2013/09/12/hibernate-orm-430-beta-4-release/

@hferentschik  Any idea what might be wrong? Do you see something wrong in the disqus id (12-09-2013-hibernate-orm-430-beta-4-release)?

hferentschik:
Took me a while to realize what you are talking about :-) Just to make sure, we are talking about the blog "summary" under the title. You basically adding a comment count link
with-count
without-count
If there is at least one comment all is good, but for posts with zero comments, nothing is rendered.

> Any idea what might be wrong? Do you see something wrong in the disqus id (12-09-2013-hibernate-orm-430-beta-4-release)?

Nope, when looking at the post it properly displays '0 comments' at the bottom, so the integration seems to work.

Otherwise a nice little gimmick.

hferentschik:
Actually the problem does not seem to be related to zero comments in general. Here for example we have multiple zero comments - http://staging.in.relation.to/emmanuel-bernard/ - and it works ok. Also, I created a test comment, but the display of comments still did not work :-(

hferentschik:
Maybe you can check with Firebug which request the underlying JavaScript tries to make. Maybe you see an error there or if not you can then try curl to do the same request. Maybe this  will show you where the problem is.

hferentschik:
Closing this pull request. Seems we need to get back to the drawing board first on this one.

PR #52, opened 2015-06-26
WEBSITE-318 Workaround for site re-generation. Instead of using the p…

hferentschik:
…age.url use page.output_path which for our purposes are identical

For some reason the page.url gets somehow re-set or maybe not set at all during re-generation. The extact cause is still unclear.
Instead rely on output_path which is always set


hferentschik:
Thanks

PR #53, opened 2015-07-13
WEBSITE-375 Hibernate Search Generic JPA introduction Blog

s4ke:
This is the .adoc version of the blog here: [snip, please read the contents from the pr]


posts/Martin/2015-07-13-introduction-to-hibernate-search-genericjpa.adoc
```diff
@@ -0,0 +1,460 @@
...
+
+In the following article I will show you how you can leverage the 
+power of Hibernate Search even though you chose not to use a 
+Hibernate based persistence provider in your application by using the new
+https://github.com/Hotware/Hibernate-Search-JPA[*Hibernate Search GenericJPA project*].
+
+== The example project
+
+Imagine a Book store (ok, a really simple one).
+Information about their books are stored in their MySQL database and accessed via EclipseLink.
```
gunnarmorling: Who is "their"?

posts/Martin/2015-07-13-introduction-to-hibernate-search-genericjpa.adoc
```diff
@@ -0,0 +1,460 @@
...
+    <modelVersion>4.0.0</modelVersion>
+    <groupId>org.hibernate</groupId>
+    <artifactId>hibernate-search-jpa-example</artifactId>
+    <version>0.2.0.0</version>
+    <packaging>jar</packaging>
+    <name>hibernate-search-jpa-example</name>
+ 
+    <build>
+        <plugins>
+            <plugin>
```
gunnarmorling: I don't think that's needed for the example? I'd minimize it as much as possible.

posts/Martin/2015-07-13-introduction-to-hibernate-search-genericjpa.adoc
```diff
@@ -0,0 +1,460 @@
...
+----
+
+[source,xml]
+----
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <groupId>org.hibernate</groupId>
+    <artifactId>hibernate-search-jpa-example</artifactId>
+    <version>0.2.0.0</version>
```
gunnarmorling: What's the reason for that uncommon versioning schema? For that example in this post I would not use the org.hibernate group id, rather com.example and 1.0.0-SNAPSHOT or so.

posts/Martin/2015-07-13-introduction-to-hibernate-search-genericjpa.adoc
```diff
@@ -0,0 +1,460 @@
...
+            <artifactId>junit</artifactId>
+            <version>4.12</version>
+            <scope>test</scope>
+        </dependency>
+ 
+         <!-- this is needed for adding search later -->
+        <dependency>
+            <groupId>org.hibernate</groupId>
+            <artifactId>hibernate-search-jpa</artifactId>
+            <version>0.2.0.0</version>
```
gunnarmorling, line 88: Why four digits? And is it available in Maven Central? Or should one add the JBoss repo to the POM?

posts/Martin/2015-07-13-introduction-to-hibernate-search-genericjpa.adoc
```diff
@@ -0,0 +1,460 @@
...
+book in the database to help the customers find what they want. 
+This is where a normal database like MySQL gets into trouble. 
+Yes, you can do queries like:
+
+[source,sql]
+----
+SELECT b FROM BOOK b WHERE b.name like ‘%Hobbit%’;
+----
+
+But these kind of queries don’t have all the power
```
gunnarmorling: these=this

posts/Martin/2015-07-13-introduction-to-hibernate-search-genericjpa.adoc
s4ke, line 88: As of now this whole project can not be found in any Maven repository.

posts/Martin/2015-07-13-introduction-to-hibernate-search-genericjpa.adoc
```diff
@@ -0,0 +1,460 @@
...
+----
+
+You may have noticed loading a properties file in this snippet.
+This is the configuration needed for Hibernate Search. Let’s take a look at it next.
+
+[source]
+----
+# we use are using MySQL. this is needed to 
+# create the triggers for updating
+org.hibernate.search.genericjpa.searchfactory.triggerSource=\
```
gunnarmorling, line 274: You should describe the general architectural approach using triggers somewhere. I only vaguely remember it, other readers won't have a clue at all of what's going on ;)

posts/Martin/2015-07-13-introduction-to-hibernate-search-genericjpa.adoc
```diff
@@ -0,0 +1,460 @@
...
+# if you are managing your Transactions by yourself
+# org.hibernate.search.genericjpa.searchfactory.transactionManagerProvider.jndi=\
+# java:jboss/TransactionManager
+org.hibernate.search.genericjpa.searchfactory.useJTATransactions=false
+ 
+# for this example we store our directory in ram.
+hibernate.search.default.directory_provider=ram
+ 
+# we could add extra properties as specified in the Hibernate Search 
+# documentation (http://docs.jboss.org/hibernate/search/5.3/reference/en-US/html_single/#_configuration) next
```
gunnarmorling, line 300: Better link to http://docs.jboss.org/hibernate/stable/search/reference/. And make it nicely rendered with a text.

posts/Martin/2015-07-13-introduction-to-hibernate-search-genericjpa.adoc
```diff
@@ -0,0 +1,460 @@
...
+ 
+import org.hibernate.search.genericjpa.annotations.Event;
+import org.hibernate.search.genericjpa.annotations.IdFor;
+import org.hibernate.search.genericjpa.annotations.Updates;
+ 
+@Entity(name = "BookUpdates")
+// what is the original table name? and what table is this entity saved to
+// this is kinda duplicated, but needed
+@Updates(originalTableName = "Book", tableName = "BookUpdates")
+@Table(name = "BookUpdates")
```
gunnarmorling, line 358: So I need a specific update table/entity for each entity? That seems quite a lot of work for the user. Two questions:
- Why does it have to be entities, couldn't your project work somehow based on tables?
- If tables suffice, could your project generate those on the fly?

Bottom line, requiring these entities seems quite a bit to ask users to create and maintain.

If this really, really is needed (I hope not), I'd do it like this annotation-wise: `@Updates(forEntity=Book.class, tableName="BookUpdates")` to avoid encoding the string-based table name another time.

s4ke, line 358: Yes I know that it is a lot of work for the user, but as of now I didn't find a way to properly generate those tables. It is definitely possible but will require a LOT of work.

Actually I just recently found out that having these entities is actually better in some rare cases, but yeah, still a lot of work.

gunnarmorling, line 358: > having these entities is actually better in some rare cases

In which cases? Some details would help.

Tbh. I think this is not a viable road.

s4ke, line 358: The thing is that the user will ALWAYS have to configure this at some point in his code. Be it at the .xml level or via extra Entities. I played around with annotations on the original entities, and with extra config files, but the Updates entities were the easiest to understand for a probable user in my opinion.

Also, the reason I encode the String based table name twice is because I have a sub-module that doesn't require any knowledge of JPA. Yeah I know, too much abstraction is bad, but I didn't want to mix this stuff too much.

I guess it _could_ be possible to figure out all of this by doing some JPA-implementor code and some magical guessing what the ids of the entities are, but this would be extremely hard to accomplish (if not impossible, depending on what every JPA implementor exposes). And this would mean more work than I could handle at the moment. :(

About the cases this would actually be more powerful: Let's say we have an abstract entity A with some classes B, C implementing it with an extra table for both. But for some reason the DBA decides he wants to store a relation between all A's and some other class D in the same table with no extra information/discriminators added. Determining what ID belonged to which class (B or C) then gets a bit tricky. JPA already does that for you. But generating JPA entities like the *Updates classes is either very very hard or not really feasible.

And on another note: I think it should be clear that a first version of this will definitely be a bit more clunky to use than the original Hibernate Search. The updating mechanism might seem easy to improve at first sight, but after many many hours of thinking about this I came to the conclusion that it is indeed not.

hferentschik, line 358: > Tbh. I think this is not a viable road.

+1 I think there is ways to much overhead involved here. Additional entities (could they btw not be generated!?), update tables, triggers, ...

I understand there is no generic JPA way, but given that there are maybe at most two more popular JPA implementations out there, would it not have made sense to write some custom integration for each of them?

s4ke, line 358: > +1 I think there is ways to much overhead involved here. Additional entities (could they btw not be generated!?), update tables, triggers, ...

The triggers are automatically created, the only thing needed are the annotations.

I've stated that before from time to time now. I am not happy with how the Update system works at the moment as well, BUT I think it's a reasonable enough default implementation for now.

It actually has some benefits when you think of it. In an application where a DB Admin doesn't allow for code to generate any database related code, the automatic generation from the info we find from the JPA provider might be problematic as 1. it's less obvious to the user what's happening under the hoods and 2. the information we can get from the JPA provider is actually quite limited. There is no easy way (as far as I can tell) that we can get information about what data-type the IDs are, so this might end up being a lot of guess-work. For example the API covering this for EclipseLink is sparsely documented at most.

I actually started writing some JPA provider specific code, but then dropped it because it was too much work to figure out every specific detail. Errors in this code would be horrendous, so I decided that this would be a feature that should be done after I get some real user-feedback.

posts/Martin/2015-07-13-introduction-to-hibernate-search-genericjpa.adoc
gunnarmorling, line 88: Hum, how do you expect the reader to run this example then?

s4ke, line 88: Well. I didn't want to put anything up in Maven central until I had some beta-testing done. Should I put this up anyways?

gunnarmorling, line 88: No, but you should provide instructions allowing the reader to actually run
your example. In this case this probably would be instructions for checking
out your project, build it from source and install into the local Maven
repo.

posts/Martin/2015-07-13-introduction-to-hibernate-search-genericjpa.adoc
gunnarmorling, line 358: > I've stated that before from time to time now. I am not happy with how the Update system works at the moment as well, BUT I think it's a reasonable enough default implementation for now.

Understood, it's only that I tend to disagree ;)

> In an application where a DB Admin doesn't allow for code to generate any database related code, the automatic generation...

That one I don't really understand. Before you said the triggers _are_ generated, so this does not add up really. IMO it should be similar to the "hbm2ddl" option: provide a way for letting your lib generate everything automatically, and allow the user to do it manually if needed by disabling this option.

> we can get information about what data-type the IDs 

Have you looked into `EntityManagerFactory#getMetaModel()`? You can get the Java types at least. From there it should be not too hard to get to the corresponding SQL column types. In the worst case you do a hard-coded mapping of Java type -> column type depending on the provider (as Hardy is saying, it's really only two others there which are of interest for now).

s4ke, line 358: For generating the entities:

I thought about it, but that's the same problem as with generating the Tables, just with some extra problems added because I have to figure out a way to dynamically load these in other JPA providers (apart from Hibernate).

Please remember: This is project is NOT intended to be used with Hibernate if you are 100% sure you will stick to it. Then, Hibernate Search is probably the better choice. But if you don't want to bind yourself with Hibernate, you can use this.

I understand that it's not as easy to use as Hibernate Search and I don't know whether that's even possible in the near future, but users that need Search capabilities and want to use EclipseLink/OpenJPA will have at least an option now.

And as complicated the Update system might seem for now, it actually has a lot of benefits even though it is clunky to use. The clunkyness however can _easily_ be improved in later versions.

s4ke, line 358: > Have you looked into EntityManagerFactory#getMetaModel()? You can get the Java types at least. From there it should be not too hard to get to the corresponding SQL column types. In the worst case you do a hard-coded mapping of Java type -> column type depending on the provider (as Hardy is saying, it's really only two others there which are of interest for now).

Yes, I have. And what about @IdClass, @EmbeddedId? Those are a different beast. For Integer, Long, etc. I guess I could force a bigint. Strings are straightforward as well, yes.

posts/Martin/2015-07-13-introduction-to-hibernate-search-genericjpa.adoc
s4ke, line 88: Okay, will do.

hferentschik, line 88: > No, but you should provide instructions allowing the reader to actually run your example. In this case this probably would be instructions for checking out your project,

+1 No point to publish maven coordinates, if the artifacts are not published.

posts/Martin/2015-07-13-introduction-to-hibernate-search-genericjpa.adoc
s4ke, line 358: Also, this code isn't that easily written... I have to map every table in the whole index tree including the collection tables. Considering _all_ the edge cases would have proved to be a nightmare.

I just thought that for a first version this would be too much work for me during this summer of code. As I have stated before, I am really unhappy with it is right now, but I wanted to have testers for the first version before I work on an improved version.

I have silently been working on this for almost 10 months now and didn't get any real user feedback. I think now it's time, even though the Update system might not be mature, yet.

gunnarmorling:
During lunch I've thought a bit more about the general approach. Obviously, any polling-based approach will always cause some delay and potentially consume CPU time without any actual changes.

So as an alternative, have you considered to invoke your library directly from within the triggers? This means, you still may do inserts with the affected ids into some staging table(s) (to not delay transactions on actual business tables), but via triggers on the staging table(s) you could then notify your lib.

In case of H2 you can call Java code directly, for many others you could invoke a simple REST interface for instance. I know Oracle, MySQL and Postgres have HTTP client facilities, I'd assume it's similar for many others.

s4ke:
@gunnarmorling : I know that some databases support those events, I actually worked with stuff like this already in an unrelated project.

But to keep things simple I don't want to do this at the moment. This would require more database specific code than I already have at the moment and other features need improving first. Even though I won't be able to do this during this summer, this is in fact a good thing to keep in mind and apart from the table generation one of the things that should be high on the priority list for future versions of this.

s4ke:
In fact it would be quite awesome to have this at one point as it removes the current async behaviour completely.

s4ke:
About the CPU consumption: I don't think that this would

gunnarmorling:
Exactly. If it were me, I'd first investigate all the different options
before going down one path. The current approach comes with many
disadvantages as you already have found out. So I wouldn't necessarily
pursue this approach if others seems more promising.

Concerning simplicity, I would even argue that simple one-line triggers
invoking a REST endpoint and the corresponding backend - again rather
trivial as it would just have to take the incoming id and entity name and
trigger indexing - may even be simpler than what you have now.

As you say it leads you down the path of DB specifics, but personally I
wouldn't mind this, given the advantages this approach may have.

s4ke:
> Exactly. If it were me, I'd first investigate all the different options
> before going down one path. The current approach comes with many
> disadvantages as you already have found out. So I wouldn't necessarily
> pursue this approach if others seems more promising.

Okay. To be honest, when I first started working on the trigger approach I didn't think about database events.

For your REST approach I would need to start a HTTP server in my application if I get your idea right. Then, what would you do if multiple programs are using this database? You would have to either change the trigger OR register more triggers (which isn't supported by every Database). Or you would store the information about the listening servers in the database.

Currently my trigger approach doesn't support multiple UpdateSources on the same table (that's what I call them), but I am planning on changing this.

Could you maybe elaborate more on such a REST system would work? Maybe I misunderstood something.

s4ke:
Also note: The async backend is the lowest common denominator that would work for every database that has trigger support.

Also, I think that in most use-cases having an async backend wouldn't be such a big problem at all. If you want your stuff to be found immediately after you insert it, you can always just index it manually for critical entities. The only problem I see here is the permanent polling which consumes too much CPU power.

s4ke:
By the way. Doing this with database events rather than with the polling has the exact same problems apart from the cpu usage. We still have to register triggers on the correct table and we have to figure out which column(s) are the ids. This goes for mapping tables as well again.

Considering this, a hybrid approach would probably be the best to use. Use triggers to store everything in extra event tables and then notify the backend. But for now I think the polling version suffices as it works with every database and implementing specific logic for every database would be too much work for now.

emmanuelbernard:
@s4ke, I did read the version you had up on blogspot and wrote some notes. Here they are verbatim. Bare in mind that they might be a bit outdated.

hibernate-search-jpa version 0.2.0.0 no need for 4 numbers

"In such an engine you have the power about how the name would be indexed"
not quite English

Why is @InIndex necessary, why isn’t @Indexed sufficient?

Can you make @DocumentId optional like it is for plain Hibernate Search

Why is it called a JPASearchFactoryController and not a JPASearchFactory?

"we have a searchfactory that is supposed to get it's updates from SQL"
-> it’s -> its and you can remove the superfluous words "like supposed" to and "we have"

Properties are prefixed with org.hibernate.search.genericjpa.searchfactory
I don’t think searchfactory is meaningful here, i’d rather see the functional area rather than a technical artifact

You cannot extract the fact that an EMF uses JTA vs resource local (i.e. the need for the true/false property)?

Can we generate these synthetic entities like envers do? I guess the problem is to have them discovered by the third party persistence provider.

What does the trigger thing is about (`triggerSource`) This is not explained.
What is org.hibernate.search.genericjpa.searchfactory.type=sql about? What are the alternatives
Why do we need a specific entity for each indexed entity, what is happening inside

No way to attach the SearchFactory lifecycle to the EMF?

The post should probably explain why we need the notion of trigger and the additional entity. As a reader it’s unclear how to glue things.

s4ke:
@emmanuelbernard : thanks.

> hibernate-search-jpa version 0.2.0.0 no need for 4 numbers

yes, should probably change this.

> "In such an engine you have the power about how the name would be indexed"
> not quite English

s/about/over ?

> Why is @InIndex necessary, why isn’t @Indexed sufficient?

I explained this somewhere above already, I think. It's needed to determine the actual class an object is of (also needed for non-root entities). For example EclipseLink sub-classes their entities so a *.getClass() won't work.

> Can you make @DocumentId optional like it is for plain Hibernate Search

It is.

> Why is it called a JPASearchFactoryController and not a JPASearchFactory?

I didn't want to confuse it with the SearchFactory term that exists in the Hibernate Search engine.

> "we have a searchfactory that is supposed to get it's updates from SQL"
> -> it’s -> its and you can remove the superfluous words "like supposed" to and "we have"

+1, thanks.

> Properties are prefixed with org.hibernate.search.genericjpa.searchfactory
> I don’t think searchfactory is meaningful here, i’d rather see the functional area rather than a 
> technical artifact

+1

> You cannot extract the fact that an EMF uses JTA vs resource local (i.e. the need for the true/false > property)?

I would need to parse all the persistence.xml files. So at the moment, no.

> Can we generate these synthetic entities like envers do? I guess the problem is to have them  discovered by the third party persistence provider.

Exactly. Also getting the right information out of the EntityManagerFactory will be a big problem. I am working on a less-boilerplate heavy approach at the moment. This will still require configuration by the user but removes the need for extra entities.

> What does the trigger thing is about (triggerSource) This is not explained.
> What is org.hibernate.search.genericjpa.searchfactory.type=sql about? What are the alternatives
> Why do we need a specific entity for each indexed entity, what is happening inside

okay. will have to clarify that.

> No way to attach the SearchFactory lifecycle to the EMF?

Haven't found a way yet. I tried delegating to the original PersistenceProvider, but that involves a lot of ClassLoader hacks and it only worked for Hibernate ORM.

> The post should probably explain why we need the notion of trigger and the additional entity. As a > reader it’s unclear how to glue things.

I will rewrite some parts of this after I changed the Update approach. Then I will only have to explain what triggers are for.

s4ke:
BTW: thanks to @gunnarmorling and @hferentschik for the input as well :)

emmanuelbernard:
> > "In such an engine you have the power about how the name would be indexed"
> >  not quite English
> 
> s/about/over ?

I would rephrase it like this.
A full-text search engine lets you decide precisely how the name value is indexed.

emmanuelbernard:
> > Can you make @DocumentId optional like it is for plain Hibernate Search
> 
> It is.

Ok, I'd remove it from the example then to make it as simple as possible.

emmanuelbernard:
> > Why is it called a JPASearchFactoryController and not a JPASearchFactory?
> 
> I didn't want to confuse it with the SearchFactory term that exists in the Hibernate Search engine.

OK, we probably need a better name. I don't think controller conveys any real meaning. Plus your builder method says it is a searchFactory `JPASearchFactoryController e = Setup.createSearchFactory(...);`

emmanuelbernard:
> > No way to attach the SearchFactory lifecycle to the EMF?
> 
> Haven't found a way yet. I tried delegating to the original PersistenceProvider, but that involves a lot of ClassLoader hacks and it only worked for Hibernate ORM.

I think I have a solution.

You will create a JPA PersistenceProvider implementation and ask people to use it.

```
<persistence xmlns="http://java.sun.com/xml/ns/persistence"
             xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
             xsi:schemaLocation="http://java.sun.com/xml/ns/persistence http://java.sun.com/xml/ns/persistence/persistence_2_0.xsd"
             version="2.0">
    <persistence-unit name="hsearchable-" transaction-type="JTA">
        <provider>org.hibernate.search.genericjpa.HibernateSearchJPAPersistence</provider>
        ...
         <properties>
            <property name="hibernate.search.genericjpa.provider" value="f.q.c.n.EclipseLinkProvider" />
           <!-- Hibernate Search properties -->
            <property name="hibernate.search.*" value="..." />
            <!-- Delegated provider properties -->
            <property name="org.eclipse.eclipselink.*" value="..." />
        </properties>
    </persistence-unit>
</persistence>
```

That way you can create an `EMF` that holds the SearchFactory instance and whatever other objects you need and delegates all methods to the underlying `EMF` of the delegated provider.

You gain:
- control of the lifecycle
- via `unwrap` a way to get access to whatever specific API the full-text use needs
- if the provider swallow all the properties it does not know, you are not affected as you come first
- you can read whether that's a JTA or resource local transaction easily
- a more integrated environment
- lots of bugs to fix in JipiJapa but your target is not WildFly so you sholud be safe ;)

emmanuelbernard:
> > The post should probably explain why we need the notion of trigger and the additional entity. As a > reader it’s unclear how to glue things.
> 
> I will rewrite some parts of this after I changed the Update approach. Then I will only have to explain what triggers are for.

I think an intro summarizing what genericjpa does at the beginning would be immensely useful as well. More like a drill down approach instead of straight into the details.

s4ke:
@emmanuelbernard: I actually tried creating my own PersistenceProvider (this is what i meant with "delegating"), in fact my approach was the exact same. EclipseLink doesn't want to be used like this. It ignores ClassLoader hacks etc. Even if I found a way for this to work with EclipseLink/OpenJPA, it would probably end up being the hackiest code ever conceived.

Besides, EJB applications can just add the hibernate-search-ejb module which bootstraps everything and lets you inject a SearchFactoryController via `@EJB`. Similar things can be done for other environments.

This worked like a charm for Hibernate ORM, but that doesn't get me anywhere for other JPA providers.

s4ke:
> OK, we probably need a better name. I don't think controller conveys any real meaning. Plus your  builder method says it is a searchFactory JPASearchFactoryController e = Setup.createSearchFactory(...);

oh, good catch about the builder method. That one must have slipped through while refactoring.

emmanuelbernard:
@s4ke I don't understand why you needed any classloading hack of any sort. What I describe was plain delegation, nothing fancy. Can you explain in more details your experiment?

s4ke:
@emmanuelbernard : Yes, can do.

I wanted to delegate to the other providers as well. I tried it out for EclipseLink by doing a PersistenceProvider that wanted to delegate all the methods of the `PersistenceProvider` interface to the equivalent for EclipseLink. However, I found out that EclipseLink did internal checking of some sorts to make sure that it doesn't return a EntityManagerFactory for a `persistence.xml` it didn't belong to (this is the expected behaviour I think). So then I tried to override the classloader used by EclipseLink to locate the `persistence.xml` and return it with the `<provider></provider>` line changed to `<provider>org.eclipse.persistence.jpa.PersistenceProvider</provider>`. However this did not work. For Hibernate ORM this Classloader hack worked like a charm as far as I remember.

And as far as I remember it didn't work for Hibernate ORM without any hack.

emmanuelbernard:
Ah ok got it. Too bad because the PP road would solve a lot of the other problems you have.

s4ke:
The only thing it would have helped me with is the lifecycle which now the user has to manage on their own. For the update system I will be dropping the need for entities alltogether and use plain SQL. Then I can at least generate the tables by code.

hferentschik:
@Sanne, what is the state on this pull request?

s4ke:
@hferentschik I guess we can drop it. I will have to rewrite it anyways for a release version.

Sanne:
ok, I'll close it then.

PR #54, opened 2015-07-16
Website 380

gunnarmorling:


emmanuelbernard:
It creates an empty `<meta name="author"/>` in edition mode but that's ok I guess.

emmanuelbernard:
Why the unique extention namespace change?

gunnarmorling:
> It creates an empty <meta name="author"/> in edition mode but that's ok I guess.

Ah, no, that's not what I wanted. That's a leftover of some forth and back, I've just force-pushed an update to this branch.

> Why the unique extention namespace change?

Just a drive-by change as I debugged the splitter; Really we should do it for all the extensions we customized.

I see you closed it but it's not merged yet? If so, re-pull before merging. Otherwise I'll push my little rework myself.

emmanuelbernard:
it was already merged, so yes you will need to push you small improvement @gunnarmorling.

Re namespace, I'm not really certain of the usefulness. For sure now we have a nicely inconsistent approach :)

Sanne:
@emmanuelbernard where did you merge it to? I'm not seeing this applied.

Sanne:
(re-opened)

PR #55, opened 2015-07-17
Add Twitter and G+ share buttons on each post

Sanne:


emmanuelbernard:
I personally do no like these kinds of social bars. Do you guys think it is useful ? G+ is way less popular than fb, where do we stop?

On more minor points, the flashy colors are at odds with our sober theme and the immensely long banner looks a bit awkward on tablets and phones.

hferentschik:
> I personally do no like these kinds of social bars. Do you guys think it is useful ?

+1 same for me

I would not do it, but I guess it is a matter of taste and attitude towards
these type of media

> On more minor points, the flashy colors are at odds with our sober theme

+1 is there a way to get icons fitting our style?

Sanne:
Since we both tweet and share on g+ our articles, this might make it more convenient for others too.
It's not useful to blog if nobody reads it..

I agree with the style issues. These are the smallest one available, I guess one could hack alternative scripts but I'm not sure if we'd be violating any brand restriction. For sure using an alternative method would be unusual, hence less effective.

sebersole:
I personally like having them.  I agree they look too awkward atm.

hferentschik:
We might not agree completely on whether or not to add these icons, but I think we agree that we don't do it in this form. Closing this pull request.

PR #59, opened 2016-01-12
jenkins build failed and local don't build for me.

maxandersen:
Can someone verify/tell me what is supposedely wrong with this new entry ?

### DIFF
--- a/posts/Max/2016-01-12-jboss-community-asylum-hibernate-strikes-back-episode-v.html.erb
+++ b/posts/Max/2016-01-12-jboss-community-asylum-hibernate-strikes-back-episode-v.html.erb
@@ -1,0 +1,33 @@
+---
+title: "JBoss Community Asylum - Hibernate Strikes Back Episode V"
+author: "Max Andersen"
+blogger_name: "Max"
+creation_date: "12-01-2016"
+original_tags: []
+tags: [JBoss Asylum]
+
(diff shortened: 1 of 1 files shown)


hferentschik:
Out of interest, why are you using erb files? The existing erb files are a result from the import of the old blog posts. It was easier for us to import HTML (aka erb files). However, new blog posts can be written in Markdown or Asciidoc. I think most of us prefer the latter. See also https://github.com/hibernate/in.relation.to#write-a-blog

hferentschik:
> local don't build for me

Just for this post or in general? If the latter, have you tried the Docker based approach - https://github.com/hibernate/in.relation.to/tree/staging/docker

hferentschik:
BTW, seems there are issues with the Jenkins setup. Hang tight ;-)

dreab8:
the problem seems related with the file name

2016-01-12-jboss-community-asylum-hibernate-strikes-back-episode-v.erb

instead of

2016-01-12-jboss-community-asylum-hibernate-strikes-back-episode-v._html_
.erb

maxandersen:
I use the .erb files because I haven't been able to get anything else to work/look remotely decent but mainly because it was faster :)

hferentschik:
Merged. I was so free to change the date to make it appear as the latest blog and I added some 'br' at the end, since your image was higher than your text column which caused some problems with the horizontal divider.

PR #60, opened 2016-01-25
Update 'level' attribute for inactive bloggers

Sanne:
The '10' values look like a mistake? however I'm not sure about the meaning of this 'level' attribute.

### DIFF
--- a/_data/people/BrettMeyer.md
+++ b/_data/people/BrettMeyer.md
@@ -1,7 +1,7 @@
 ---
 name: Brett Meyer
 photo:
-level: 10
+level: 1
 website: http://3riverdev.com
 location: Fort Wayne, IN, USA
 occupation: Software Engineer @ Red Hat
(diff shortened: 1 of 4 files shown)


DavideD:
The level is used to order the list of authors that appear on the website; they get ordered first by level and then by name.

I wouldn't touch it unless you need to make some changes to the way the list is rendered.

gunnarmorling:
I had left Strong and Shane enabled as they had posted not that long ago. +1 for disabling them, as I don't think they'll write new stuff. Scott hasn't posted in a long time, but I considered him part of the team.  

Regarding 1 vs 10, it doesn't change anything in this specific case, as Brett comes first by alphabet but also by comparing 10 to 1 descending :) But good to change it for the sake of consistency. Everything > 0 is shown (see blogger_list.html.haml).

Sanne:
thanks for all explanations. So ok to merge it?

gunnarmorling:
Yes, please go ahead. I'd probably leave Scott, but your call :)

Sanne:
Ok I'll leave him.. deserves to be part of the team, my reasoning was just that he hasn't blogged in 5 years so users don't have much value on clicking on his name..

PR #62, opened 2016-03-23
Change tag 'Off-topic' to 'Off topic'

DavideD:
@emmanuelbernard it seems that this post is using the wrong tag. It stated in the readme that we should use "Off topic", I've replaced it


emmanuelbernard:
thanks, just apply

DavideD:
Merged https://github.com/hibernate/in.relation.to/pull/62

PR #67, opened 2016-09-30
Introducing Hibernate Search Sort DSL

yrodiere:
http://staging.in.relation.to/2016/09/30/introducing-hibernate-search-sort-dsl/

### DIFF
--- a/posts/Sanne/2016-10-06-TripleHibernateSearchRelease.adoc
+++ b/posts/Sanne/2016-10-06-TripleHibernateSearchRelease.adoc
@@ -1,0 +1,54 @@
+= Triple Hibernate Search release
+Sanne Grinovero
+:awestruct-tags: [ "Hibernate Search", "Elasticsearch", "Releases" ]
+:awestruct-layout: blog-post
+---
+
+Today we have three releases of Hibernate Search!
+
+I'm proud to announce our team is a bit larger nowadays, and more contributors are volunteering too, so we managed to increase the development pace. Today we release version `5.6.0.Beta3`, `5.7.0.Alpha1` and `5.5.5.Final`.
+
+Version 5.6.0.Beta3:: the latest version of our main development branch, with experimental Elasticsearch integration.
+
+Version 5.7.0.Alpha1:: essentially the same as `5.6.0.Beta3`, but compatible with Hibernate ORM version `5.2.x`.
+
+Version 5.5.5.Final:: a maintenance release of our stable branch.
+
+== A 5.7 preview released when 5.6 isn't out yet?
+
+Let me explain this unusual decision was taken to accomodate for the needs of you all.
+
+The `5.6` series is creating a lot of anticipation with the Elasticsearch integration being a very welcome new feature; it's meant to be an experimental new feature as we won't break our APIs yet while all integration needs are analyzed, still it's taking a bit longer than expected and even though it's and experimental feature we don't want to rush it and need to finish it up properly.
+
+In the meantime the Hibernate ORM project released a series `5.2.x`, and several users have been asking to get an Hibernate Search version compatible with it. We could not upgrade our `5.6` series yet, as then people using an older Hibernate ORM would not be able to play with the Elasticsearch integration.
+
+So now that `5.6` is in good shape - we decided the next release will be a candidate release - we felt we could already publish a `5.7` version, which is just exactly the same but in a new branch made compatible with the very latest Hibernate ORM.
+
+== How is the Elasticsearch integration coming?
+
+It's maturing at high speed. The biggest obstacles have been resolved, so we definitely look out for more feedback at this point; as mentioned, the next version will be a candidate release.
+
+Hibernate Search now has a proper *Sorting API*: watch this space as we'll publish a dedicated blog about it, or get a peek at the
+https://docs.jboss.org/hibernate/search/5.6/reference/en-US/html_single/#query-sorting[query sorting] paragraph in the documentation.
+
+This is an important milestone, as it makes sorting queries on Elasticsearch possible through our DSL.
+
+== How to get these releases
+
+All versions are available on Hibernate Search's http://hibernate.org/search/[web site].
+
+Ideally use a tool to fetch it from Maven central; these are the coordinates:
+
+====
+[source, XML]
+----
+<dependency>
+   <groupId>org.hibernate</groupId>
+   <artifactId>hibernate-search-orm</artifactId>
+   <version>5.6.0.Beta3</version>
+</dependency>
+----
+====
+
+Downloads from https://sourceforge.net/projects/hibernate/files/hibernate-search/[Sourceforge] are available as well.
+


posts/Yoann/2016-09-30-introducing-hibernate-search-sort-dsl.adoc
```diff
@@ -0,0 +1,88 @@
...
+:awestruct-tags: [ "Hibernate Search" ]
+:awestruct-layout: blog-post
+---
+With Elasticsearch support coming as a technological preview in Hibernate Search 5.6, you would think we're leaving out other features. Well, think again! Enters the Sort DSL, which will work with Elasticsearch of course, but also with the good ol' Lucene backend.
+
+The point here is to provide an API to build sort descriptions easily, without knowing everything about Hibernate Search features added on top of Lucene, such as `DistanceSortField`. And while we're at it, we're making it a modern, fluid API.
+
+== Overview
+
+The API works pretty much like the query builder. The query builder is the starting point, actually: a new `sort()` method has been added to the `QueryBuilder` interface. Here's what it looks like:
```
DavideD: I would just keep the sentence:

> A new `sort()` method has been added to the `QueryBuilder` interface:

Everything else seems redundant.

Additional point if you can avoid the passive form, something like the QueryBuilder interface now has an additional sort() method ...`

posts/Yoann/2016-09-30-introducing-hibernate-search-sort-dsl.adoc
```diff
@@ -0,0 +1,88 @@
...
+----
+  .sort()
+    .byDistance()
+      .onField("location")
+      .fromLatitude(24.0d)
+      .andLongitude(32.0d)
+    .createSort()
+----
+ 
+== Advanced use
```
DavideD: Instead of collect the following features in a unique section, why not to put them in their own section?

posts/Yoann/2016-09-30-introducing-hibernate-search-sort-dsl.adoc
```diff
@@ -0,0 +1,88 @@
...
+[source, JAVA]
+----
+  .sort()
+    .byField("publishingDate_sort").desc() // Most recently published first 
+      .onMissingValue().sortFirst() // Not published yet => put this upper on the list
+    .andByField("custom_id_sort") // Default for the case when multiple books have no publishing date
+  .createSort()
+----
+
+And of course, if you're using Elasticsearch, you may want to take advantage of that brand-new feature that appeared in the last snapshot, the one you just spotted this morning and that would really save you of a lot of trouble on your project. But, hey, the Hibernate Search team is not on the same time zone, and even if they're providing fast support, you're not getting the feature pushed into Hibernate Search in time to meet your deadline. Which is this evening, by the way.
```
DavideD: I would put this in its own section and remove the initial part: `And of course, if you're using Elasticsearch,`

DavideD: I think it's a very good point and deserve to be highlighted more

posts/Yoann/2016-09-30-introducing-hibernate-search-sort-dsl.adoc
```diff
@@ -0,0 +1,88 @@
...
+    .byNative("authors.name", "{'order':'asc', 'mode': 'min'}")
+    .andByField("title")
+  .createSort()
+----
+
+== Next...
+
+Of course, one could point out that it's not really backend-independent. The API itself mostly is, but the returned type (`Sort`) is clearly bound to Apache Lucene.
+
+Well, one day at a time: the API in its current form *can* be adapted the be completely agnostic, so it's paving the way to http://hibernate.org/search/roadmap/#hibernate-search-6-x[Hibernate Search 6.x], while still requiring no change to any other contract such as `FullTextQuery.setSort(Sort)`. And that means it will available directly in 5.6.0.Beta3!
```
gsmet: can be adapted the agnostic -> can be adapted _to_ be _backend-_agnostic?

posts/Yoann/2016-09-30-introducing-hibernate-search-sort-dsl.adoc
```diff
@@ -0,0 +1,88 @@
...
+
+So be sure to check it out, and feel free to contact us for any question, problem or simply for feedback!
```
gsmet: for feedback -> to give us your feedback?

posts/Yoann/2016-09-30-introducing-hibernate-search-sort-dsl.adoc
```diff
@@ -0,0 +1,88 @@
...
+----
+
+And of course, if you're using Elasticsearch, you may want to take advantage of that brand-new feature that appeared in the last snapshot, the one you just spotted this morning and that would really save you of a lot of trouble on your project. But, hey, the Hibernate Search team is not on the same time zone, and even if they're providing fast support, you're not getting the feature pushed into Hibernate Search in time to meet your deadline. Which is this evening, by the way.
+
+Well, guess what: you can use that feature anyway. The sorting API also allows using native sorts by passing the JSON description of this sort, which will be added to the Elasticsearch query as is:
+
+[source, JAVA]
+----
+  .sort()
+    .byNative("authors.name", "{'order':'asc', 'mode': 'min'}")
```
gsmet: @Sanne @yrodiere we should decide if we choose the 2 parameters form or the one parameter form for byNative before publishing the blog post.

posts/Yoann/2016-09-30-introducing-hibernate-search-sort-dsl.adoc
```diff
@@ -0,0 +1,88 @@
...
+----
+  .sort()
+    .byField("title")
+    .andByIndexOrder()
+  .createSort()
+----
+
+That way, if there are two books with the same title in your index, they will always keep the same relative order from one query to another.
+
+The score sort is also available with `byScore()`. Obviously, there's one key difference with that one: the sort is descending by default. Getting the least relevant entities first wouldn't be very useful.
```
DavideD: Instead of point out the obvious I would explain how you can change the order of the score..
Something like:
`If you need the least relevant results, fear not, we got you covered,  with the byScore().desc()` method.`

DavideD:
@yrodiere Great post.
Overall, I think it would be nicer to create separate sections for the most common cases, spatial and native; instead of a more general `overview` section.

My 2 cents, feel free to pick what you like.

posts/Yoann/2016-09-30-introducing-hibernate-search-sort-dsl.adoc
gsmet: When I first read that, it wasn't clear to me that the "feature appeared in the last snapshot" was a feature of Elasticsearch and not a feature of Hibernate Search (which doesn't make sense at all). Maybe "in its last snapshot" would be better?

posts/Yoann/2016-09-30-introducing-hibernate-search-sort-dsl.adoc
DavideD: Maybe we can mention that we are still deciding between to options and this might change before the final release

posts/Yoann/2016-09-30-introducing-hibernate-search-sort-dsl.adoc
yrodiere: Done, thanks.
About the passive form, is it because it's bad English or for another reason? Just curious.

posts/Yoann/2016-09-30-introducing-hibernate-search-sort-dsl.adoc
yrodiere: Well, actually it's `byScore.asc()`. By ascending score, so the lesser scores first. But yes, I changed the sentence to get to the point (while still stating the obvious, because it might not be so obvious depending on your point of view).

posts/Yoann/2016-09-30-introducing-hibernate-search-sort-dsl.adoc
yrodiere: The idea was that the "advanced use" section is mainly here to show there's more, not to explain everything else. The reference documentation will be there for people who want extensive information.

But if we want to give the native sort more visibility, then yes, there's not point in keeping that section labelled as "Advanced use".

I changed that.

posts/Yoann/2016-09-30-introducing-hibernate-search-sort-dsl.adoc
yrodiere: Right, done.

yrodiere: I changed that to also hint at the fact that this could apply to the Lucene backend, too. I'm not covering this more extensively, since it's less likely to be of use with the Lucene backend. I also addressed the possible confusion in the section title, just in case.
Does it seem clearer to you?

posts/Yoann/2016-09-30-introducing-hibernate-search-sort-dsl.adoc
yrodiere: @gsmet @DavideD I was expecting us to merge https://github.com/hibernate/hibernate-search/pull/1139 _before_ I publish this blog post anyway. I will fix this snippet before publishing if necessary.
I wouldn't want to confuse user by advertising a feature that does not even has its code released on a branch of the official repo.

posts/Yoann/2016-09-30-introducing-hibernate-search-sort-dsl.adoc
yrodiere: Right, fixed.

posts/Yoann/2016-09-30-introducing-hibernate-search-sort-dsl.adoc
yrodiere: Yeah, not sure it's correct English, but it felt more natural... You're right, better be safe.

yrodiere:
@DavideD @gsmet Thanks for your feedback, I changed the post to fix what you. Feel free to ping me if it's not satisfying. If it is, I'll push that when https://github.com/hibernate/hibernate-search/pull/1139 has been merged.

About the post layout, I wanted to first show we're addressing the most common cases (`byScore`, `byDistance`, and so on) and then show that we're not providing only basic features, without falling in the trap of a reference documentation in a blog post. That was the reason for the "Advanced use" section. I changed it a bit to highlight the native sort part a bit more.

posts/Yoann/2016-09-30-introducing-hibernate-search-sort-dsl.adoc
DavideD: Active form is easier to read and to understand. Usually, sentences become shorter and the subject of the paragraph become clearer.

It's more of a general guideline, it isn't necessarily bad form. I've found this article and it seems to have good points about when to use it: https://wordcounter.net/blog/2016/09/14/102274_how-to-identify-passive-voice.html 

I'm sure other people will feel differently about it.

posts/Yoann/2016-09-30-introducing-hibernate-search-sort-dsl.adoc
DavideD: I prefer to keep them separated because it makes section shorter and from the title you can have an overview of the new features.

Sanne:
It seems phrased as if you wanted to release it before having the API released. I think we should first release Beta3, and only then (immediately after?) release the dedicated blog. WDYT?

I think we're good to release Beta3 immediately.

yrodiere:
@Sanne Releasing it after having the API released seems better. Before releasing beta3, maybe we should merge this simple one, though: https://github.com/hibernate/hibernate-search/pull/1169

Sanne:
Ok.

PR #71, opened 2016-12-20
Blog post for HSearch 5.6.0.CR1 and 5.7.0.Beta2

yrodiere:
Any suggestions are welcome! There is not much to say here, though, because we mainly fixed bugs.

@DavideD , I don't know if we should advertise the new dynamic mapping option now. Even if it works well, like you said it's still a prototype and we may still introduce breaking changes. Plus, we did not have time to update the documentation about the limitations that actually disappeared thanks to this option. Do you think I should mention it anyway?


DavideD:
@yrodiere I would briefly mention it.
How would you expect people to leave feedback if they are not aware of it? I would just highlight that it is a preview and still in active development. Also it comes from user feedback and we should thank him for the help.

posts/Yoann/2016-12-20-hibernate-search-5-6-0-CR1-and-5-7-0-Beta2.adoc
```diff
@@ -0,0 +1,99 @@
...
+   the automatic discovery of new nodes in the Elasticsearch cluster for client-side loadbalancing and (limited) failover.
+ * https://hibernate.atlassian.net/browse/HSEARCH-2219[HSEARCH-2219]: analyzer definitions are now automatically
+   translated and added to the Elasticsearch settings when Elasticsearch indexes are created.
+   Please note there are still some limitations, most notably the definitions are not updated automatically (no support
+   for the MERGE schema management strategy).
+   See the https://docs.jboss.org/hibernate/search/5.6/reference/en-US/html_single/#elasticsearch-mapping-analyzer[documentation]
+   for more information about analyzers with Elasticsearch.
+ * ... and much more, mainly bug fixes.
+   The full change log can be found on https://hibernate.atlassian.net/secure/ReleaseNote.jspa?projectId=10061&version=25900[our JIRA instance]
+   or on https://github.com/hibernate/hibernate-search/blob/5.6.0.CR1/changelog.txt[our GitHub repository].
```
DavideD: Personally I would just pick one of the two links. In OGM , we started to point to GitHub.

posts/Yoann/2016-12-20-hibernate-search-5-6-0-CR1-and-5-7-0-Beta2.adoc
```diff
@@ -0,0 +1,99 @@
...
+
+== When will this 5.6.0 be released?
+
+As far as we are concerned, 5.6.0 is ready.
```
DavideD: I find it weird that you use 5.6.0 as version, if you prefer to be precise you should use 5.6.0.Final otherwise I think 5.6 would be enough.

DavideD:
I think it's missing  a  paragraph about how to contact us with the links to the forum, HipChat room and tag on stackoverflow

posts/Yoann/2016-12-20-hibernate-search-5-6-0-CR1-and-5-7-0-Beta2.adoc
```diff
@@ -0,0 +1,99 @@
...
+----
+<dependency>
+   <groupId>org.hibernate</groupId>
+   <artifactId>hibernate-search-elasticsearch</artifactId>
+   <version>5.6.0.CR1</version>
+</dependency>
+----
+====
+
+(Change the version to `5.7.0.Beta2` in order to test the Elasticsearch integration within Hibernate Search 5.7)
```
DavideD: Shouldn't you move this line where you are explaining how to add 5.7.0.Beta2?

DavideD: ah no, forget about it

DavideD: I would remove the parenthesis.

DavideD:
Looks good to me overall, I just left of couple of remarks.

Mainly I think you should add a section about how to contact us and be a bit more consistent when you mention the versions. I would use the format 5.7 when you don't need to be overly specific and 5.7.0.Final when you have to.

gsmet (dismissed):
I agree with Davide's comment and don't have anything else to add.

For the contact part, see what we do for HV here: https://github.com/hibernate/in.relation.to/blob/staging/posts/Guillaume/2016-12-08-hibernate-validator-540-beta1-and-534-final-out.adoc#feedback-issues-ideas

yrodiere:
Thanks! I fixed the issues. I kept the JIRA link to the changelog, though, not the GitHub one, because the latter lacks hyperlinks.

PR #74, opened 2017-01-27
Add a module to generate json feed from our posts for a given set of tags

gsmet:
Related to https://hibernate.atlassian.net/projects/WEBSITE/issues/WEBSITE-484 .

@vladmihalcea @DavideD  I wrote a small Awestruct module to generate a JSON file for a set of defined tags (don't know what we currently use for the others page though).

To be able to use them from a remote site, we should probably use JSONP but most of the work is already done.

@vladmihalcea Could you take it over from there?

Thanks!

### DIFF
--- a/_ext/json_feed_generator.rb
+++ b/_ext/json_feed_generator.rb
@@ -1,0 +1,94 @@
+require 'awestruct/context_helper'
+
+module InRelationTo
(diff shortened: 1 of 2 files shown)


gsmet:
Merged.

PR #78, opened 2017-02-13
Small post about Duration validation

marko-bekhta:
Good time of day to you @gunnarmorling  and @gsmet :)

I know that the previous post is still pending (I cleaned the other one a bit and moved the examples around as you've asked), but I was thinking that this one can be useful and interesting. As we can show others how we decided to validate `Duration` and get some feedback from users, so we will know what do they think of it.

I've tried to be as brief as possible on this one :)

Also it should probably go after the first one as it is missing "author" part and will fail to build correctly without it.

### DIFF
--- a/posts/MarkoB/2017-02-12-validator-duration-validation.adoc
+++ b/posts/MarkoB/2017-02-12-validator-duration-validation.adoc
@@ -1,0 +1,93 @@
+= "Extended support for java.time"
+Marko Bekhta
+:awestruct-tags: [ "Hibernate Validator", "Discussions" ]
+:awestruct-layout: blog-post
+---
+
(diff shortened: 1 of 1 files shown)


gunnarmorling:
Thanks, @marko-bekhta. Question is whether you want to mention `@DurationMax/Min` in the HV 6 Alpha1 release announcement, @gsmet? Do we then still need a separate follow-up post? Personally I think it'd be better handled (just showing the usage) in the release announcement. WDYT?

gsmet:
It's indeed part of the release announcement.

marko-bekhta:
@gunnarmorling @gsmet well yes, if there's a place for it in HV6 release announcement than definitely  better put it there. If you don't have it ready yet, and example in here is good enough - you can use it :)

gsmet:
Closing, it was integrated in the HV6 release announcement.

PR #114, opened 2018-05-25
[OGM] Hibernate OGM 5.4.0.Beta1 release post

DavideD:


posts/Davide/2018-05-25-hibernate-ogm-5-4-Beta1-released.adoc
```diff
@@ -0,0 +1,84 @@
...
+
+All the changes are described in detail in the
+https://hibernate.atlassian.net/secure/ReleaseNote.jspa?projectId=10160&version=31632[release notes].
+
+== Hibernate ORM 5.3
+
+You can find all the details about it in the
+http://in.relation.to/2018/05/14/hibernate-orm-530-final-release/[Hibernate ORM 5.3.0.Final release post].
+The main things I will mention here is that it comes with support for JPA 2.2 and many
+performance improvements.
```
gsmet: > and many performance improvements

I don't think that's true. There are improvements on the memory front but I don't think there's been progress on the performances.

posts/Davide/2018-05-25-hibernate-ogm-5-4-Beta1-released.adoc
```diff
@@ -0,0 +1,84 @@
...
+
+== Infinispan Remote queries
+
+You can execute queries in Infinispan using the string query language called
```
gsmet: > string query language

I would remove the "string" here, it looks weird.

posts/Davide/2018-05-25-hibernate-ogm-5-4-Beta1-released.adoc
DavideD: Ok, I will remove it
I guess I've been a bit too optimistic and the improvements side

posts/Davide/2018-05-25-hibernate-ogm-5-4-Beta1-released.adoc
```diff
@@ -0,0 +1,84 @@
...
+====
+
+With Ickle you can also write queries using the Lucene syntax, as long as the field is
+indexed. Note that this will only work if the cache already exists on the server and the
+fields are indexed. Hibernate OGM won't keep track of indexing information when it creates
+new caches.
+
+== Where can I get Hibernate OGM?
+
+You can include in your project the dialect of your choice using this Maven coordinates:
```
gsmet: s/this/these/

posts/Davide/2018-05-25-hibernate-ogm-5-4-Beta1-released.adoc
DavideD: ok, the example should be enough

posts/Davide/2018-05-25-hibernate-ogm-5-4-Beta1-released.adoc
```diff
@@ -0,0 +1,84 @@
...
+* https://discourse.hibernate.org/c/hibernate-ogm[User forum]
+* https://hibernate.atlassian.net/browse/OGM[Issue tracker]
+* http://lists.jboss.org/pipermail/hibernate-dev/[Mailing list]
+* http://stackoverflow.com[Stack Overflow]: we monitor the tag _hibernate-ogm_
+* https://www.hipchat.com/gXEjW5Wgg[HipChat]: Hibernate OGM hipchat room
+
+== Contributions
+
+Some of the new features have been contributed by 
+Sergey Chernolyas, Viet Nguyen.
```
gsmet: s/, /and/
Isn't it The Viet?

posts/Davide/2018-05-25-hibernate-ogm-5-4-Beta1-released.adoc
```diff
@@ -0,0 +1,83 @@
+= "Hibernate OGM 5.4.0.Beta1 release"
+Davide D'Alto
+:awestruct-tags: [ "Hibernate OGM", "Releases" ]
+:awestruct-layout: blog-post
+---
+
+http://hibernate.org/ogm/releases/5.4/#get-it[Hibernate OGM 5.4.0.Beta1] has been released!
+
+The main differences from the previous release is the upgrade to Hibernate ORM 5.3.0.Final, 
+the support for Infinispan Remote native queries and the use of cluster counters
```
fax4ever, line 10: JPQL and native. I think not just native

DavideD: I mentioned it with more details later.

Technically you could query using JPQL even before, it's the way we do it that's changed.

DavideD: I will add `JPQL (without using Hibernate Search)`

posts/Davide/2018-05-25-hibernate-ogm-5-4-Beta1-released.adoc
```diff
@@ -0,0 +1,83 @@
...
+for local caches when generating sequences with Infinispan Embedded.
```
fax4ever, line 11: Maybe for local cache **too** (for remote it was already provided)

DavideD: I don't think it's necessary to add the `too`, the fact the we added support for one case doesn't mean it's not already supported for other use cases.

DavideD:
Thanks, I'm going to publish it

posts/Davide/2018-05-25-hibernate-ogm-5-4-Beta1-released.adoc
```diff
@@ -0,0 +1,83 @@
...
+[source, JAVA]
+----
+String ickleQuery = "from HibernateOGMGenerated.Registry where level > 3 order by start"
+List result = session.createNativeQuery( ickleQuery )
+				.addEntity( Employee.class )
+				.list();
+----
+====
+
+With Ickle you can also write queries using the Lucene syntax, as long as the field is
```
fax4ever, line 40: Maybe we should use Ickle syntax instead of Lucene syntax. I think @Sanne is going to fix it.

posts/Davide/2018-05-25-hibernate-ogm-5-4-Beta1-released.adoc
```diff
@@ -0,0 +1,83 @@
...
+indexed. Note that this will only work if the cache already exists on the server and the
+fields are indexed. Hibernate OGM won't keep track of indexing information when it creates
```
fax4ever, line 42: At the moment it is not so easy define indexes, because indexes is defined in protoschema file. Currently protoschema file is generated by OGM, the way to override it it is not so easy and not documented,

DavideD: OK, but if the cache is defined already on the server we can query it using ickle, isn't it? THat's what I think it's important

DavideD: But feel free to prove me wrong, I'm not a super expert in this. Maybe we need to add a test case to double check. Also we might want to check if  the error is clear when the user tries to run a full-text query and that's not allowed (because the field is not indexed). @fax4ever can you create an issues for these if it make sense to you?

PR #124, opened 2018-11-29
[HSEARCH] 6.0.0.Alpha1 release

yrodiere:
Available on staging: http://staging.in.relation.to/2018/11/29/hibernate-search-6-0-0-Alpha1/

### DIFF
--- a/posts/Yoann/2018-11-29-hibernate-search-6-0-0-Alpha1.adoc
+++ b/posts/Yoann/2018-11-29-hibernate-search-6-0-0-Alpha1.adoc
@@ -1,0 +1,229 @@
+= Get a sneak peek into the future of Hibernate Search with 6.0.0.Alpha1!
+Yoann Rodière
+:awestruct-tags: [ "Hibernate Search", "Lucene", "Elasticsearch", "Releases" ]
+:awestruct-layout: blog-post
+---
+
+We just published Hibernate Search 6.0.0.Alpha1,
+the first release for the next major version of Hibernate Search.
+
+This version is not suitable for production due to its limitations and to the fact APIs are still unstable.
+However, it's a great way to get a feeling of where Hibernate Search is going and give us your feedback.
+
++++<!-- more -->+++
+
+== Getting started
+
+If you want to dive right in, a good starting point is the
+https://docs.jboss.org/hibernate/search/6.0/reference/en-US/html_single/#getting-started[getting started guide]
+included in the reference documentation.
+
+== What's new
+
+For a full list of changes since the previous releases,
+please see the https://hibernate.atlassian.net/secure/ReleaseNote.jspa?projectId=10061&version=31657[release notes].
+Be warned, though: there is a lot in there. You may find the summary below easier to process.
+
+=== Preview status
+
+Hibernate Search 6.0 is currently a *technology preview* and is not ready for production.
+
+While the core of Hibernate Search 6 is there, there are still a lot of limitations:
+
+* Features are partially implemented or simply missing.
+For example, you cannot index a `Double` property yet,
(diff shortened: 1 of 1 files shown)


Sanne:
Always over-cautious. There's no need to use bold letters to warn this is not meant for production use :)

posts/Yoann/2018-11-29-hibernate-search-6-0-0-Alpha1.adoc
```diff
@@ -0,0 +1,179 @@
...
+
+While the core of Hibernate Search 6 is there, there are still a lot of limitations:
+
+* Features are partially implemented or simply missing.
+For example, you cannot index a `Double` property yet,
+you cannot configure custom Lucene directories,
+you cannot ignore bridges in the predicate DSL,
+and so on.
+* APIs are not considered stable yet.
+* The low-level parts of Hibernate Search 5 that made its efficiency haven't been
```
Sanne: "made its efficiency" ?

Sanne (approved):
Awesome :+1:

posts/Yoann/2018-11-29-hibernate-search-6-0-0-Alpha1.adoc
```diff
@@ -0,0 +1,179 @@
+= Get a sneak peek into the future of Hibernate Search with 6.0.0.Alpha1!
+Yoann Rodière
+:awestruct-tags: [ "Hibernate Search", "Lucene", "Elasticsearch", "Releases" ]
+:awestruct-layout: blog-post
+---
+
+We just published Hibernate Search 6.0.0.Alpha1,
+the first release for the next major version of Hibernate Search.
+
+This version is **not suitable for production** due to its limitations and to the fact APIs are still unstable.
```
fax4ever: I don't know whether it makes sense, but I would like to point out also the positive elements of this new series. Such as the fact that the Hibernate Search v6 is more modular, rational and maintainable then the previous ones. So we expect in the future more efficiency and usability, due to the effort sustained here.

posts/Yoann/2018-11-29-hibernate-search-6-0-0-Alpha1.adoc
yrodiere: "made it efficient" maybe?

fax4ever: what about `... Search 5 that enhances (or improves) efficiency ..`

posts/Yoann/2018-11-29-hibernate-search-6-0-0-Alpha1.adoc
yrodiere: That's absolutely right, but I'm not sure I want to say this in a blog post targeting users. It could send implicit messages that I don't want to send:

* "Search 5 is not maintainable": it is maintainable since we're maintaining it.
* "They're breaking everything just for the beauty of it": we're not, there are legitimate reasons to break APIs.
* "Search 6 is going to be blazing fast": we haven't run performance tests yet, there's no way to tell for sure.
* "Search 6 will be bring us new features at unprecedented pace": it depends first and foremost on how many people work on Search. So far it looks good, but I don't want to over-promise.

I think what I'm trying to say is: the redesign has great potential, but I'm not sure it's something we can use to "promise" anything... Let the results speak for themselves :)

posts/Yoann/2018-11-29-hibernate-search-6-0-0-Alpha1.adoc
yrodiere: "improve its efficiency" it is. Thanks

PR #127, opened 2019-02-06
Replace links to HipChat with the appropriate links to Zulip

yrodiere:
@Sanne @DavideD @fax4ever the only links I could find were in your blog posts.

I don't want to change your blog posts without your consent, so... Are you ok with my changes?


fax4ever:
+1 and thanks for having done that

yrodiere:
Merged. Please ping me if you disagree with some changes, I'll revert as necessary.

PR #156, opened 2021-04-30
Hibernate Reactive 1.0.0.CR4 release post

DavideD:
Also available on staging: https://staging.in.relation.to/2021/04/30/hibernate-reactive-1_0_0_CR4/


gavinking:
Great!

DavideD:
Fixed the syntax

DavideD:
@gavinking is the post good enough?

gavinking:
I would say:

> It’s now possible to use Hibernate Reactive with **both** Hibernate ORM 5.4 and Hibernate ORM 5.5.

Or explicitly clarify that we previously only supported 5.4.

Also I would clarify that no code changes are needed for people using `withSession()` as recommended.

DavideD:
Ah, good point. I'm going to eat something and then I will update it

DavideD:
Merged!
Thanks

PR #163, opened 2022-03-08
Retire OGM

yrodiere:


gsmet (approved):

_partials/header.html.haml
```diff
@@ -9,9 +9,10 @@
     %a.item{:href => "#{site.hibernate_org.base_url}orm/"} ORM
     %a.item{:href => "#{site.hibernate_org.base_url}search/"} Search
     %a.item{:href => "#{site.hibernate_org.base_url}validator/"} Validator
-    %a.item{:href => "#{site.hibernate_org.base_url}ogm/"} OGM
-    %a.item{:href => "#{site.hibernate_org.base_url}tools/"} Tools
+    - if real_page.project == 'ogm'
+      %a.item{:href => "#{site.hibernate_org.base_url}ogm/"} OGM
```
DavideD, line 13: I don't understand this change. Why can't you just remove it?

yrodiere: :shrug: That's what Guillaume did on hibernate.org, so I did the same here.

gsmet: Basically, we display the menu if people are in the OGM section already.

We do the same for the website.

DavideD: I kinda get it why we would do that in hibernate.org.
But why do we need it for in.relation.to? If you click on any button  of the header it will go to `hibernate.org`

DavideD: By the way, I miss this in hibernate.org, but `OGM` should probably be after `Others`. This way we can think of it as a sub-voice of `Others`: Others > OGM

PR #164, opened 2022-04-29
[ORM] Blog post on Boolean type type mapping in ORM 6

jrenaat:
Signed-off-by: Jan Schatteman <jschatte@redhat.com>

### DIFF
--- a/posts/Jan/2022-04-29-mapping-boolean-types-in-hibernate-orm-6.adoc
+++ b/posts/Jan/2022-04-29-mapping-boolean-types-in-hibernate-orm-6.adoc
@@ -1,0 +1,172 @@
+= Mapping Boolean attributes in ORM 6
+Jan Schatteman
+:awestruct-tags: [ "Hibernate ORM" ]
+:awestruct-layout: blog-post
+:icons: font
+:url-hibernate-base: https://docs.jboss.org/hibernate/orm/6.0
+:url-hibernate-javadoc: {url-hibernate-base}/javadocs
+:url-hibernate-ug: {url-hibernate-base}/userguide/html_single/Hibernate_User_Guide.html
+:url-sdk11-javadoc: https://docs.oracle.com/en/java/javase/11/docs/api
+
+
+By default, Hibernate ORM maps a Boolean value to a BOOLEAN column. However, not all databases have a built-in BOOLEAN datatype and not all applications will want to use BOOLEAN (legacy schema e.g.) .
+
+Hibernate provides multiple ways to map Boolean value to other SQL types; this post explores the most common ways.
+
+.Implicit Boolean mapping
+====
+[source, JAVA, indent=0]
+----
+
+    @Basic
+    private boolean aBoolean;
+
+----
+====
+How Hibernate chooses to map a boolean implicitly is determined by a series of cascading checks:
+
+1. It first checks whether the link:{url-hibernate-javadoc}/org/hibernate/cfg/AvailableSettings.html#PREFERRED_BOOLEAN_JDBC_TYPE[hibernate.type.preferred_boolean_jdbc_type] setting has been specified and, if so, uses it;
+2. Next it consults the link:{url-hibernate-javadoc}/org/hibernate/dialect/Dialect.html#getPreferredSqlTypeCodeForBoolean()[Dialect] to see if it reports a preferred SQL type and, if so, uses it;
+3. Otherwise it will use link:{url-sdk11-javadoc}/java.sql/java/sql/Types.html#BOOLEAN[Types.BOOLEAN]
+
+This default, 'implicit' mapping can be modified in several ways.
+
+== Using Jakarta Persistence AttributeConverter
+An `AttributeConverter` provides a natural way to map a boolean accounting for different storage representations.
+
+Hibernate provides 3 built-in `AttributeConverter` implementations for the most common cases:
+
+* *YesNoConverter* encodes a boolean value as 'Y' or 'N',
+* *TrueFalseConverter* encodes a boolean value as 'T' or 'F', and
+* *NumericBooleanConverter* encodes the value as an integer, 1 for true, and 0 for false.
+
+[[convert]]
+.Using built-in AttributeConverters
+====
+[source, JAVA, indent=0]
+----
+// this will get mapped to CHAR or NCHAR
+@Basic
+@Convert(converter = org.hibernate.type.YesNoConverter.class)
+boolean convertedYesNo;
+
+// this will get mapped to CHAR or NCHAR
+@Basic
+@Convert(converter = org.hibernate.type.TrueFalseConverter.class)
+boolean convertedTrueFalse;
+
+// this will get mapped to TINYINT
+@Basic
+@Convert(converter = org.hibernate.type.NumericBooleanConverter.class)
+boolean convertedNumeric;
+----
+====
+
+If another type of conversion is needed, one can also implement a custom `AttributeConverter`. See the link:{url-hibernate-ug}#basic-jpa-convert[User Guide] for details about using `AttributeConverter`.
+
+.Defining a custom AttributeConverter
+====
+[source, JAVA, indent=0]
+----
+
+    @Converter
+    public class CustomBooleanConverter implements AttributeConverter<Boolean, Character> {
+        ...
+    }
+
+----
+====
+
+In the above <<convert,examples>>, the converter will only be applied to the specific annotated attribute.
+
+Usually, however, this storage representation will be the same across an entire application or organization.  In such cases, annotating each and every boolean attribute individually is tedious and error prone.
+
+Jakarta Persistence defines a portable way to auto-apply a converter to all boolean types found in the domain model using `jakarta.persistence.Converter#autoApply`.
+
+[[convert-auto]]
+.Auto-applied AttributeConverter
+====
+[source, JAVA, indent=0]
+----
+@Converter( autoApply = true )
+public class CustomBooleanConverter implements AttributeConverter<Boolean, Character> {
+    ...
+}
+----
+====
+
+See the link:{url-hibernate-ug}[User Guide] for details about such 'auto-applied' converters.
+
+How would we define such auto-application for the aforementioned built-in `AttributeConverters` though? One option is to explicitly tell Hibernate about it using `orm.xml`:
+
+.<converter/>
+====
+[source, XML, indent=0]
+----
+<entity-mappings>
+    ...
+        <converter>org.hibernate.type.YesNoConverter</converter>
+    ...
+</entity-mappings>
+----
+====
+
+== @ConverterRegistration
+The imminent Hibernate 6.1 release will provide an alternative solution for this, by registering converters using the new `@ConverterRegistration` annotation.
+This annotation is a corollary to the `orm.xml` `<converter/>` element.
+
+Consider Hibernate's built-in `YesNoConverter`, and let's say we want to auto-apply it.  The converter itself is not defined for auto-application (and that would not be appropriate given potential discovery problems). However, we can use this new annotation, which can be specified at the class or package level  (it's scope is global regardless), as shown below:
+
+.@ConverterRegistration
+====
+[source, JAVA, indent=0]
+----
+@ConverterRegistration( converter=org.hibernate.type.YesNoConverter.class, autoApply=true )
+----
+====
+
+`ConverterRegistration` will be covered in more detail in the 6.1 version of the User Guide.
+
+
+== Hibernate extensions
+
+=== `@JdbcTypeCode / @JdbcType`
+
+If one wants to map a specific boolean attribute to something different from the default for the current database, then the easiest way to achieve this is to annotate that attribute with the desired `@JdbcTypeCode`, or with `@JdbcType` and the desired `JdbcType` implementor. In the latter case one can either use an implementation provided by Hibernate that accounts for the desired transformation, or provide a custom implementation.
+
+.Apply a specific Jdbc Type locally to a Boolean
+====
+[source, JAVA, indent=0]
+----
+    // this Boolean attribute will be converted to a CHAR in the database
+    @JdbcTypeCode(SqlTypes.CHAR)
+    private Boolean toCharBoolean;
+
+    // this Boolean attribute will be converted to a SMALLINT in the database
+    @JdbcType( SmallIntJdbcType.class )
+    private Boolean toSmallIntBoolean;
+----
+====
+
+If the default Boolean mapping to the database needs to change on an application-wide basis, then one would use the `@JdbcTypeRegistry` annotation as shown below:
+
+.Apply a specific Jdbc Type globally to all Boolean attributes
+====
+[source, JAVA, indent=0]
+----
+
+    @Entity
+    @JdbcTypeRegistration(registrationCode = Types.BOOLEAN, value = SmallIntJdbcType.class)
+    public static class AnEntity {
+        @Id
+        private Integer id;
+
+        private Boolean someBooleanAttribute;
+
+        //Getters and setters omitted for brevity
+    }
+----
+====
+
+Don't hesitate to check out the link:{url-hibernate-ug}#basic[User Guide] for more examples and further reading!
+
--- a/posts/Andrea/2022-05-25-hibernate-orm-602-final.adoc
+++ b/posts/Andrea/2022-05-25-hibernate-orm-602-final.adoc
@@ -1,0 +1,45 @@
+= "Hibernate ORM {released-version} released"
+Andrea Boriero
+:awestruct-tags: ["Hibernate ORM", "Releases"]
+:awestruct-layout: blog-post
+:released-version: 6.0.2.Final
+:docs-url: https://docs.jboss.org/hibernate/orm/6.0
+:migration-guide-url: {docs-url}/migration-guide/migration-guide.html
+:user-guide-url: {docs-url}/userguide/html_single/Hibernate_User_Guide.html
+:release-id: 32070
+
+Today, we published a new maintenance release of Hibernate ORM 6.0: {released-version}.
+
+== What's new
+
+This release introduces a few minor improvements as well as bug fixes.
+
+=== Optimisation
+
+We have improved how the `LazyAttributeLoadingInterceptor` is dealing with lazy fields, this makes its processing a bit more efficient and saves a bit of memory as well (see https://hibernate.atlassian.net/browse/HHH-15274[HHH-15274]).
+
+=== Bugfixes
+
+You can find the full list of changes in this version https://hibernate.atlassian.net/secure/ReleaseNote.jspa?version={release-id}&styleName=Html&projectId=10031[here].
+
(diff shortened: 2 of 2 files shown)


posts/Jan/2022-04-29-mapping-basic-custom-type-in-hibernate-orm-600.adoc
```diff
@@ -0,0 +1,319 @@
+= Custom basic type mapping in Hibernate 6
+:icons: font
```
sebersole: There is additional metadata you need for a blog post -

```
:awestruct-tags: [ "Hibernate ORM" ]
:awestruct-layout: blog-post
```

I'd also maybe change the title.  Something like "Basic Vaue Mapping in ORM 6"

posts/Jan/2022-04-29-mapping-basic-custom-type-in-hibernate-orm-600.adoc
```diff
@@ -0,0 +1,319 @@
...
+Jan Schatteman <jschatte@redhat.com>
+
+This is a blog post about the basic type mapping system in Hibernate 6, which has changed quite a bit with respect to previous versions, specifically in how individual parts that make up a basic type mapping are configured. Before 6.0, the user could implement and register a custom org.hibernate.type.BasicType contract, but this is now considered deprecated.
```
sebersole:
```suggestion
Much has changed in ORM 6.0 with regard to its type system.  

The https://docs.jboss.org/hibernate/orm/6.0/userguide/html_single/Hibernate_User_Guide.html#basic[User Guide]
does a good job covering the mapping of basic values.  Migrating users will also find the
https://docs.jboss.org/hibernate/orm/6.0/migration-guide/migration-guide.html[Migration Guide] helpful.  This post
digs a little deeper into the effects of those changes on the handling and mapping of basic values.  


Historically Hibernate provided 3 ways to customize the mapping of a basic value -

* `UserType`
* `BasicType`
* `AttributeConverter`


6.0 removes the ability to customize mappings through a custom `BasicType`, but adds a new "compositional" approach.

Let's use the following entity as an example for these approaches:

// you model here


// from here, I'd break it down into sections -

== UserType

// mostly the same - only reading method changed (name to position)


== AttributeConverter

// completely the same


== Compositional

// here is the meat


```

sebersole: Not sure why it kept the "This is a blog post about the basic type mapping system in Hibernate 6, ..." part in my suggestion :D

jrenaat: ehhh, probably because you left it in there, either that or GH already has AI sniffing for outstanding prose in place :rofl:

posts/Jan/2022-04-29-mapping-boolean-types-in-hibernate-orm-6.adoc
```diff
@@ -0,0 +1,145 @@
...
+====
+[source, JAVA, indent=0]
+----
+
+    @Basic
+    private boolean aBoolean;
+
+----
+====
+In the above example the Boolean attribute will be mapped to what is returned by the call to `JdbcTypeIndicators.getPreferredSqlTypeCodeForBoolean()`, which in turn reverts to either the value configured for the `AvailableSettings.PREFERRED_BOOLEAN_JDBC_TYPE` setting or, if that is not set, the value returned from the call to `Dialect.getPreferredSqlTypeCodeForBoolean()`, the default being `Boolean` in case a specific dialect doesn't override this.
```
sebersole: Stuff like `JdbcTypeIndicators.getPreferredSqlTypeCodeForBoolean()` is way to low-level for users.

The discussion here needs to be much more generalized IMO.  Something like -

```
How Hibernate chooses to map a boolean implicitly is determined by a series of cascading checks:

  1. It first checks whether the link:{javadoc-AvailableSettings}#PREFERRED_BOOLEAN_JDBC_TYPE[hibernate.type.preferred_boolean_jdbc_type] setting has been specified and, if so, uses it;
  2. Next it consults the link:{javadoc-Dialect}#getPreferredSqlTypeCodeForBoolean[Dialect] to see if it reports a preferred SQL type and, if so, uses it;
  3. Otherwise it will use link:{javadoc-Types}#BOOLEAN[Types.BOOLEAN]
```

posts/Jan/2022-04-29-mapping-boolean-types-in-hibernate-orm-6.adoc
```diff
@@ -0,0 +1,149 @@
...
+----
+
+    @Basic
+    private boolean aBoolean;
+
+----
+====
+How Hibernate chooses to map a boolean implicitly is determined by a series of cascading checks:
+
+1. It first checks whether the link:{javadoc-AvailableSettings}#PREFERRED_BOOLEAN_JDBC_TYPE[hibernate.type.preferred_boolean_jdbc_type] setting has been specified and, if so, uses it;
```
sebersole: You have to define these ;)

e.g.

```
:javadoc: https://docs.jboss.org/hibernate/orm/6.0/javadocs/
:javadoc-AvailableSettings: {javadoc}/org/hibernate/cfg/AvailableSettings.html
```

posts/Jan/2022-04-29-mapping-boolean-types-in-hibernate-orm-6.adoc
```diff
@@ -0,0 +1,149 @@
...
+
+NOTE: in impending ORM 6.1 release, a new `@ConverterRegistration` was added to make this much easier. It allows modifying certain aspects of already defined `AttributeConverters`, such as the application domain (`autoApply`) or change the data type the conversion should be applied on to another compatible data type (e.g. a subclass of the original data type).
+
+.Customizing built-in AttributeConverters
+====
+[source, JAVA, indent=0]
+----
+// this will apply the converter to all Boolean attributes
+@Basic
+@ConverterRegistration( converter = org.hibernate.type.YesNoConverter.class, autoApply = true )
```
sebersole: This is not allow on fields or methods.  Only packages and types (classes, etc).

posts/Jan/2022-04-29-mapping-boolean-types-in-hibernate-orm-6.adoc
```diff
@@ -0,0 +1,149 @@
...
+=== `@JdbcTypeCode / @JdbcType`
+
+If one wants to map a specific boolean attribute to something different from the default for the current database, then the easiest way to achieve this is to annotate that attribute with the desired `@JdbcTypeCode`, or with `@JdbcType` and the desired `JdbcType` implementor. In the latter case one can either use an implementation provided by Hibernate that accounts for the desired transformation, or provide a custom implementation.
+
+.Apply a specific Jdbc Type locally to a Boolean
+====
+[source, JAVA, indent=0]
+----
+    // this Boolean attribute will be converted to a CHAR in the database
+    @JdbcTypeCode(SqlTypes.CHAR)
```
sebersole, line 142: Have you verified that this actually works?

posts/Jan/2022-04-29-mapping-boolean-types-in-hibernate-orm-6.adoc
```diff
@@ -0,0 +1,149 @@
...
+        <class>CustomBooleanConverter</class>
+        ...
+    </persistence-unit>
+----
+====
+
+
+== Hibernate extensions
+
+=== `@JavaType`
```
sebersole: I'd just get rid of this.

Could additionally mention `UserType`

posts/Jan/2022-04-29-mapping-boolean-types-in-hibernate-orm-6.adoc
```diff
@@ -0,0 +1,149 @@
...
+        private Integer id;
+
+        private Boolean someBooleanAttribute;
+
+        //Getters and setters omitted for brevity
+    }
+----
+====
+
+Check out the link:{user-guide-url}#basic[User Guide] for more examples.
```
sebersole: You'll also need to define this `user-guide-url` variable.

posts/Jan/2022-04-29-mapping-boolean-types-in-hibernate-orm-6.adoc
```diff
@@ -0,0 +1,149 @@
+= Mapping Boolean attributes in ORM 6
+:awestruct-tags: [ "Hibernate ORM" ]
+:awestruct-layout: blog-post
+:icons: font
+Jan Schatteman <jschatte@redhat.com>
+
+By default, Hibernate maps a Boolean value to a BOOLEAN column. However, not all databases have a built-in BOOLEAN datatype and not all applications will want to use BOOLEAN (legacy schema e.g.).
```
DavideD: Maybe I'm nitpicking, but are we going to use `Hibernate` as synonym for `Hibernate ORM`?

sebersole: Yep, we often do.  For most users, Hibernate == Hibernate ORM

posts/Jan/2022-04-29-mapping-boolean-types-in-hibernate-orm-6.adoc
jrenaat: Believe it or not, but yes, I actually have ;)
What makes you doubt it wouldn't, btw?

sebersole: This tells Hibernate which JDBC type to use, not how to perform a conversion from boolean -> char.

What gets stored when you do this?

jrenaat: in H2: CHAR(1) CHECK ("B" IN('N', 'Y'))
So, Hibernate does magic even beyond it's makers' knowledge ... :D

posts/Jan/2022-04-29-mapping-boolean-types-in-hibernate-orm-6.adoc
jrenaat: UGH, how silly, copy-paste overlook. Will correct

posts/Jan/2022-04-29-mapping-boolean-types-in-hibernate-orm-6.adoc
jrenaat: I had no idea, thanks for the pointer

posts/Jan/2022-04-29-mapping-boolean-types-in-hibernate-orm-6.adoc
jrenaat: Changed it to Hibernate ORM all the same, it's a fair point IMO.

posts/Jan/2022-04-29-mapping-boolean-types-in-hibernate-orm-6.adoc
```diff
@@ -0,0 +1,163 @@
...
+====
+How Hibernate chooses to map a boolean implicitly is determined by a series of cascading checks:
+
+1. It first checks whether the link:{url-hibernate-javadoc}/org/hibernate/cfg/AvailableSettings.html#PREFERRED_BOOLEAN_JDBC_TYPE[hibernate.type.preferred_boolean_jdbc_type] setting has been specified and, if so, uses it;
+2. Next it consults the link:{url-hibernate-javadoc}/org/hibernate/dialect/Dialect.html#getPreferredSqlTypeCodeForBoolean()[Dialect] to see if it reports a preferred SQL type and, if so, uses it;
+3. Otherwise it will use link:{url-sdk11-javadoc}/java.sql/java/sql/Types.html#BOOLEAN[Types.BOOLEAN]
+
+This default, 'implicit' mapping can be modified in several ways.
+
+== JPA compliant, using AttributeConverter
```
sebersole: No need for `JPA compliant, ` here.  It adds no value.

I'd also condense this whole conversation into something like:


```
== AttributeConverter

Using an `AttributeConverter` is a natural way to map a boolean accounting for different storage representations.  
See the link:{url-hibernate-ug}#basic-jpa-convert[User Guide] for details about using `AttributeConverter`.

Hibernate provides 3 built-in `AttributeConverter` implementations for the most common cases:

YesNoConverter:: encodes boolean values as 'Y' or 'N'
TrueFalseConverter:: encodes boolean values as 'T' or 'F'
NumericBooleanConverter:: encodes boolean values as an integer, 1 for true and 0 for false


[[convert]]
.Using AttributeConverter
====
[source, JAVA, indent=0]
----
@Entity
class TheEntity {
    ...

    // this will get mapped to CHAR or NCHAR
    @Basic
    @Convert(converter = org.hibernate.type.YesNoConverter.class)
    boolean convertedYesNo;

    // this will get mapped to CHAR or NCHAR
    @Basic
    @Convert(converter = org.hibernate.type.TrueFalseConverter.class)
    boolean convertedTrueFalse;

    // this will get mapped to TINYINT
    @Basic
    @Convert(converter = org.hibernate.type.NumericBooleanConverter.class)
    boolean convertedNumeric;

    // use our custom converter
    @Basic
    @Convert(converter = CustomBooleanConverter.class)
    boolean convertedCustom;
}

@Converter
public class CustomBooleanConverter implements AttributeConverter<Boolean, Character> {
    ...
}
----
====


== Auto-applied AttributeConverter

In the above <<convert,examples>>, the conversions will only be applied to the annotated attribute.

Usually, however, this storage representation will be the same across an entire application or organization.  In such cases, annotating each and every boolean attribute individually is tedious and error prone.

Jakarta Persistence defines a portable way to auto-apply a converter to all boolean types found in the domain model using `jakarta.persistence.Converter#autoApply`.  


[[convert-auto]]
.Auto-applied AttributeConverter
====
[source, JAVA, indent=0]
----
@Converter( autoApply = true )
public class CustomBooleanConverter implements AttributeConverter<Boolean, Character> {
    ...
}
----
====

// We should actually move this discussion about how Hibernate becomes aware of auto-applied 
// converters into the User Guide.  I just noticed we don't really have any content there about that.
// And then:

See the link:{user-guide-url}[User Guide] for details about auto-applied converters.



[[convert-reg]]
== ConverterRegistration

Starting with the imminent 6.1 release, Hibernate also supports registering converters using the new `@ConverterRegistration` annotation.  This annotation is a corollary to the `orm.xml` `<converter/>` element.

Consider Hibernate's built-in `YesNoConverter`, and let's say we want to auto-apply it.  The converter itself is not defined for auto-application (and that would not be appropriate given potential discovery problems).

One option is to explicitly tell Hibernate about it using `orm.xml`:


.<converter/>
====
[source, JAVA, indent=0]
----
<entity-mappings>
    ...
        <converter>org.hibernate.type.YesNoConverter</converter>
    ...
</entity-mappings>
----
====


Another is to use this new annotation, which can be specified at the class or package level.  It's scope is global regardless.

.@ConverterRegistration
====
[source, JAVA, indent=0]
----
@ConverterRegistration( converter=org.hibernate.type.YesNoConverter.class, autoApply=true )
----
====

`ConverterRegistration` will be covered in more detail in the 6.1 version of the User Guide.


// == Hibernate Extensions
// === `@JdbcTypeCode / @JdbcType`


== `@JdbcTypeCode / @JdbcType`

```

jrenaat: Seriously though Steve, it's one thing to correct obvious errors and/or omissions and suggest improvements, and another to rewrite the whole thing. At one point or another you're gonna have to either accept the fact that if someone else wrote something it's not gonna be 100% like you would have done it, or just write it yourself.

sebersole: You are new to writing public blogs.  You asked me for help.  I am suggesting that you write way too much content for a blog post and how you can make the content simpler.  If that irks you, sure.. I'll do it

jrenaat: It doesn't irk me, I just feel maybe you're overthinking this a bit. But sure, we can keep chipping away at this until it's reduced to nothing.

jrenaat: Ok, I've given my own 'twist' to your suggestions.
For one, I don't see the need for an Auto-applied AttributeConverter section, after all it's about attributeconverters with just an attribute set to true, imo a section of its own suggests that those are some kind of special entities in their own right.
I did not know that (or forgot about) putting a converter in orm.xml auto-applies it, so obviously that has to go in there.

posts/Jan/2022-04-29-mapping-boolean-types-in-hibernate-orm-6.adoc
beikov: Might be worth mentioning in the blog post, that `Y`/`N` is the "default" representation as per `BooleanJavaType`

posts/Jan/2022-04-29-mapping-boolean-types-in-hibernate-orm-6.adoc
```diff
@@ -0,0 +1,172 @@
...
+----
+====
+
+See the link:{url-hibernate-ug}[User Guide] for details about such 'auto-applied' converters.
+
+How would we define such auto-application for the aforementioned built-in `AttributeConverters` though? One option is to explicitly tell Hibernate about it using `orm.xml`:
+
+.<converter/>
+====
+[source, JAVA, indent=0]
```
beikov:
```suggestion
[source, XML, indent=0]
```

beikov (commented):
Personally, I would love to see a some kind of "conclusion" that gives a "recommendation" maybe (given that we can provide that). I think that with converters, literals might not work as expected i.e. the following code:

```java
CriteriaBuilder cb = session.getCriteriaBuilder();
CriteriaQuery cq = session.createQuery( TestEntity.class );
cq.where( cb.equal( root.get( "convertedBoolean" ), cb.literal( true ) ) );
```

but we'd have to check to make sure. It would be nice though, if it doesn't work as expected (rendering as literal to SQL), that we "warn" users about this and hence provide guidance.

posts/Jan/2022-04-29-mapping-boolean-types-in-hibernate-orm-6.adoc
sebersole: > YesNoConverter:: encodes boolean values as 'Y' or 'N'

sebersole:
> literals might not work as expected

They *should*.  At least from HQL.  Fairly certain there are tests for that.  But more tests always ok :D

yrodiere:
Hey @jrenaat @sebersole @beikov, should we close this or do you still plan to merge it someday?

yrodiere:
> Hey @jrenaat @sebersole @beikov, should we close this or do you still plan to merge it someday?

Bump.

yrodiere:
Ok I'm pretty sure nobody cares by now... Closing, feel free to reopen :)

PR #165, opened 2022-06-23
[ORM] Add article about new 6.1 features

beikov:


yrodiere (commented):
@beikov Looks nice! Maybe you could link to this post from the page dedicated to ORM 6.1 on hibernate.org? (https://hibernate.org/orm/releases/6.1/ / https://github.com/hibernate/hibernate.org/blob/production/orm/releases/6.1/index.adoc )
Or even add a bit more details there, as you see fit :)

PR #167, opened 2023-02-07
Fix links containing `issues/\?jql` plus `%20` and fix even more uses of `ReleaseNote.jspa` in links to Jira

yrodiere:
For some reason `issues/?jql=project=PROJECTID%20AND%20fixVersion=VERSIONID` does not work well on Firefox, at least on our Jira instance. `issues/?jql=project=PROJECTID+AND+fixVersion=VERSIONID` works, though.

Also, I had apparently missed quite a few uses of `ReleaseNote.jspa` that relied on Asciidoc variables (`{releaseId}`) that my regexp didn't catch.


dreab8 (approved):
LGTM

yrodiere:
Merged, thanks

PR #237, opened 2025-09-25
Sponsor footer/Security policy link update

marko-bekhta:
I don't remember if we were supposed to update those or not ... so just in case 😃 feel free to close if not 😃


gavinking:
It would make sense to update the logo to an IBM logo I guess.

marko-bekhta:
I think we need a 🟢 from IBM legal + from Commonhaus to add it .... but yeah, that's where I'm unsure what the steps are 😃

yrodiere:
> I think we need a 🟢 from IBM legal + from Commonhaus to add it .... but yeah, that's where I'm unsure what the steps are 😃

This. Please don't add anything for IBM -- or any company, for that matter -- without consulting their legal department.

yrodiere:
I forgot: Regarding IBM, we have the contacts already. Work in progress.

yrodiere:
@marko-bekhta why did you close this though? It seems fine

PR #264, opened 2026-05-15
small blog post showing the impact of what some small changes can hav…

stalep:
…e on performance

### DIFF
--- a/posts/Stale/2026-05-29-hibernate-json-dirty-checking-performance.adoc
+++ b/posts/Stale/2026-05-29-hibernate-json-dirty-checking-performance.adoc
@@ -1,0 +1,302 @@
+= Two Hibernate Annotations That Cut Our JSON Processing Time in Half
+Ståle Pedersen
+:awestruct-tags: [ "Hibernate ORM", "Performance", "JSON" ]
+:awestruct-layout: blog-post
+:toc:
+:toclevels: 2
+---
+
+When we profiled our application's data import pipeline, we expected the bottleneck to be in our business logic -- JQ filters, JavaScript evaluation, recursive SQL queries. Instead, *Hibernate's internal JSON handling consumed over 50% of CPU time* and allocated 20 GB of unnecessary objects during a routine import of 5 records.
+
+Two small annotation changes fixed it. No code changes, no architectural redesign -- just telling Hibernate what it needs to know about how our application uses its data.
+
+== The Application
+
+https://github.com/Hyperfoil/h5m[h5m] is a lightweight rewrite of https://github.com/Hyperfoil/Horreum[Horreum], a performance regression detection system. It accepts JSON benchmark results, transforms them through a DAG (directed acyclic graph -- a pipeline of dependent computation steps where each node's output feeds into downstream nodes) and detects regressions using statistical methods. The core entity stores computed results as JSONB:
+
+[source,java]
+----
+@Entity
+public class ValueEntity extends PanacheEntity {
+
+    @Column(columnDefinition = "JSONB")
+    @JdbcTypeCode(SqlTypes.JSON)
+    @Basic(fetch = FetchType.LAZY)
+    public JsonNode data;
+
+    // ... other fields
+}
+----
+
+Each upload creates dozens of `ValueEntity` instances through a work queue -- JQ filter results, fingerprints, change detection outputs. A typical import processes thousands of these entities per minute.
+
+== Profiling: Where Is the Time Going?
+
+We profiled a 5-record import from a production dataset using https://github.com/async-profiler/async-profiler[async-profiler] with CPU sampling:
+
+[source,bash]
+----
+java -agentpath:libasyncProfiler.so=start,event=cpu,file=profile.jfr,jfr \
+     -jar app.jar import-data ...
+----
+
+The results were surprising:
+
+[cols="3,1,1"]
+|===
+| Category | CPU Samples | %
+
+| *Hibernate dirty-checking*
+| 6,087
+| *34%*
+
+| GC
+| 4,397
+| 25%
+
+| *Hibernate deepCopy*
+| 2,361
+| *13%*
+
+| JIT compilation
+| 2,084
+| 12%
+
+| Actual computation
+| 1,187
+| 7%
+
+| Hibernate JSON serialize
+| 507
+| 3%
+|===
+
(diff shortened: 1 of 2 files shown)


beikov (approved):
From a first rough read this looks good to me. Anything specific you want me to review?

mbellade (approved):
This is great! +100 on discouraging blind `Cascade.MERGE` usage!

posts/Stale/2026-05-15-hibernate-json-dirty-checking-performance.adoc
```diff
@@ -0,0 +1,300 @@
...
+Ståle Pedersen
+:awestruct-tags: [ "Hibernate ORM", "Performance", "JSON" ]
+:awestruct-layout: blog-post
+:toc:
+:toclevels: 2
+---
+
+When we profiled our application's data import pipeline, we expected the bottleneck to be in our business logic -- JQ filters, JavaScript evaluation, recursive SQL queries. Instead, *Hibernate's internal JSON handling consumed over 50% of CPU time* and allocated 20 GB of unnecessary objects during a routine import of 5 records.
+
+Two small annotation changes fixed it. No code changes, no architectural redesign -- just telling Hibernate what it should already know about our data.
```
yrodiere: > telling Hibernate what it should already know about our data

I wonder how Hibernate could possibly "already know" that a mutable class is effectively immutable in the context of a given application...

Do you mean "what it needs to know about our data"?

posts/Stale/2026-05-15-hibernate-json-dirty-checking-performance.adoc
```diff
@@ -0,0 +1,300 @@
...
+@Entity
+public class Work extends PanacheEntity {
+
+    @ManyToMany(cascade = {CascadeType.PERSIST, CascadeType.MERGE}, fetch = FetchType.LAZY)
+    @JoinTable(name = "work_values", ...)
+    public List<ValueEntity> sourceValues;
+}
+----
+
+When we called `em.merge(work)` to persist a new `Work` entity, Hibernate cascaded the merge to every `ValueEntity` in `sourceValues`. Each of these was an *already-managed, unmodified* entity -- but the cascade caused Hibernate to re-serialize their JSONB columns for UPDATE dirty-checking anyway.
```
yrodiere: This seems to imply that without the cascade merge, the dirty-checking would be skipped... but if an entity is attached to a session, we need to dirty-check it; that's precisely how Hibernate can tell it's unmodified. So I don't see how removing the cascade merge would remove the need to dirty-check.
Am I misunderstanding this?

Maybe you mean the cascade merge was causing a *second* dirty checking? I could see that happening, but it looks like an opportunity for optimization...

posts/Stale/2026-05-29-hibernate-json-dirty-checking-performance.adoc
```diff
@@ -0,0 +1,300 @@
...
+|===
+
+*47% of CPU was spent on Hibernate's internal JSON dirty-checking and snapshot management* -- not on our application logic, not on database queries, but on Hibernate comparing JSON columns to decide whether they changed.
+
+== The Problem: FormatMapperBasedJavaType.deepCopy
+
+When Hibernate manages an entity with a JSON column (`@JdbcTypeCode(SqlTypes.JSON)` with a `JsonNode` field), it needs to detect changes for dirty-checking. On every `em.flush()`, Hibernate:
+
+1. *Deep-copies* the JSON value by serializing it to a string and deserializing it back (`FormatMapperBasedJavaType.deepCopy`) -- this creates the snapshot
+2. *Compares* the current value against the snapshot using `ObjectNode.equals()` -- a recursive tree comparison
```
yrodiere, line 81: This looks... oddly complex. I started https://hibernate.zulipchat.com/#narrow/channel/132094-hibernate-orm-dev/topic/JSON.20dirty.20checking/with/595739983 to ask some (probably naive) questions about it.

yrodiere: One question arised from that discussion: if your nodes are not mutable, why aren't you simply using a `String` typed attribute, which would make the whole conversation about mutability irrelevant?

posts/Stale/2026-05-15-hibernate-json-dirty-checking-performance.adoc
stalep: Yes, I meant that Hibernate was not given enough info about the data structure. I'll fix it.

posts/Stale/2026-05-15-hibernate-json-dirty-checking-performance.adoc
stalep: Yes, so this was change would not have been impactful without the immutable change above. I'll change the text to make this clearer.

posts/Stale/2026-05-29-hibernate-json-dirty-checking-performance.adoc
stalep: Thats because data is frequently accessed as JsonNode throughout the codebase. I do not think this is uncommon.

yrodiere: I see; so you do care about the structure -- if that's the case, as @beikov suggested, you might be interested in [embeddable mapped as json/structs](https://docs.hibernate.org/orm/7.3/userguide/html_single/#embeddable-mapping-aggregate). It's a bit easier to manipulate on the application side, and if you are able to model the *whole* data like this (without relying on `JsonNode`), you bypass the mutability/dirty-check problem completely -- @beikov might be able to confirm?

stalep: Thanks for the pointer to aggregate embeddable mappings, I wasn't aware of that feature. It doesn't fit h5m's current design though as it's intentionally schemaless (we can't force everyone to have the same benchmark result format :). Different node types in the DAG pipeline produce completely different JSON structures. A jq filter might output {"cpu": 45}, a fingerprint node produces {"testName": "perf"}, and a detection node produces {"value": 95.3, "bound": 90.0, "direction": "above"}. The output shape depends on user-defined jq/JS/jsonpath expressions, so there's no single @Embeddable class that could represent all of them.

PR #269, opened 2026-07-01
Quick followup showing impact of configuring 2LC correctly

stalep:


posts/Stale/2026-08-19-hibernate-immutable-2lc-9x-speedup.adoc
```diff
@@ -0,0 +1,359 @@
...
+""")
+    .setParameter("rootId", root.id)
+    .setParameter("nodeId", node.id)
+    .getResultList();
+
+List<ValueEntity> values = new ArrayList<>(ids.size());
+for (Number id : ids) {
+    ValueEntity v = em.find(ValueEntity.class, id.longValue());
+    if (v != null) values.add(v);
+}
```
mbellade, line 117: It would be much better to use the relatively new `Session.findMultiple()` API, that can optimize the lookup of multiple IDs in the database by running a single batched query instead of (potentially) 1 database query per ID if no cache is available.

```suggestion
List<ValueEntity> valueEntities = session.findMultiple(ValueEntity.class, ids);
```

Note this API is Hibernate specific as of 7.x, but it will also be available in the JPA standard `EntityManager` interface from Hibernate 8.0 (JPA 4).

posts/Stale/2026-07-01-hibernate-immutable-2lc-9x-speedup.adoc
```diff
@@ -0,0 +1,359 @@
...
+
+Three annotations work together, each telling Hibernate something specific about the data:
+
+=== @Immutable on the entity
+
+Tells Hibernate to skip dirty-checking entirely. No snapshot creation on load, no comparison at flush time. The entity is read-only from Hibernate's perspective.
+
+=== @Cache(usage = CacheConcurrencyStrategy.READ_ONLY)
+
+This is where the cache optimization happens. With `READ_ONLY`, Hibernate switches from `StandardCacheEntryImpl` to `ReferenceCacheEntryImpl`.
```
mbellade: This is not strictly true: there are more parameters at play when determining whether we can use reference caching, first of all the user must request is explicitly via: [hibernate.cache.use_reference_entries](https://docs.hibernate.org/orm/current/userguide/html_single/#settings-hibernate.cache.use_reference_entries). The entity must also be immutable, and have no associations:

https://github.com/hibernate/hibernate-orm/blob/b49d933299f797c5d9e7bbd19c5c353e3699c6f9/hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractEntityPersister.java#L987-L1008

posts/Stale/2026-07-01-hibernate-immutable-2lc-9x-speedup.adoc
```diff
@@ -0,0 +1,359 @@
...
+Our `ValueEntity` has a `sources` relationship (the DAG edges) that was originally `FetchType.LAZY`:
+
+[source,java]
+----
+@ManyToMany(cascade = { CascadeType.PERSIST }, fetch = FetchType.EAGER)
+@JoinTable(name = "value_edge", ...)
+public List<ValueEntity> sources;
+----
+
+With the L2 cache, cached entities can be returned outside the original Hibernate session. Accessing a lazy collection on a cached entity throws `LazyInitializationException` because there is no session to initialize the proxy.
```
mbellade: This is only true for reference caching, normally 2LC entries are re-assembled into the current session and lazy-loading just works as expected.

mbellade (commented):
Hey @stalep, left a few comments hope that helps.

posts/Stale/2026-08-19-hibernate-immutable-2lc-9x-speedup.adoc
```diff
@@ -0,0 +1,359 @@
...
+    .setParameter("rootId", root.id)
+    .setParameter("nodeId", node.id)
+    .getResultList();
+----
+
+[source,java]
+----
+// AFTER: ID-only query + em.find() hits L2 cache
+List<Number> ids = em.createNativeQuery("""
+    WITH RECURSIVE sourceRecursive (v_id) AS (...)
```
beikov, line 105: FYI, HQL supports recursive CTEs, so you wouldn't even have to do two queries. You can rewrite the SQL in HQL  and avoid the `find` calls altogether.

posts/Stale/2026-07-01-hibernate-immutable-2lc-9x-speedup.adoc
```diff
@@ -0,0 +1,359 @@
...
+By telling Hibernate the truth about our data -- that it is immutable -- we allow it to make fundamentally different optimization decisions:
+
+[source,java]
+----
+@Entity
+@Immutable
+@Cacheable
+@Cache(usage = CacheConcurrencyStrategy.READ_ONLY)
+public class ValueEntity extends PanacheEntity {
+    @Column(columnDefinition = "JSONB")
```
beikov: Don't do that. JSONB on PostgreSQL is the default. We shouldn't promote the use of `columnDefinition`

```suggestion
    @Column
```

posts/Stale/2026-07-01-hibernate-immutable-2lc-9x-speedup.adoc
```diff
@@ -0,0 +1,359 @@
...
+}
+public Serializable disassemble(T value, ...) {
+    return toString( value );  // serialize to String
+}
+public T assemble(Serializable cached, ...) {
+    return fromString( (CharSequence) cached );  // parse from String
+}
+----
+
+Every `deepCopy`, `disassemble`, and `assemble` call triggers a full JSON serialization or parse. For a large JSONB column, this is devastating.
```
beikov: Can you please share the stack traces that show where `deepCopy` is used when dealing with an `@Immutable` entity? I would say that there is no need to do `deepCopy` for such entities, but need to see how this is triggered to understand what is going on.

posts/Stale/2026-07-01-hibernate-immutable-2lc-9x-speedup.adoc
```diff
@@ -0,0 +1,359 @@
...
+especially for JSON/LOB fields where the default plan performs full serialization round-trips.
+
+=== Why READ_ONLY, not NONSTRICT_READ_WRITE?
+
+Our first attempt with `@Immutable` used `NONSTRICT_READ_WRITE` as the cache strategy. This immediately failed:
+
+[source]
+----
+java.lang.AssertionError
+  at NonStrictReadWriteEntityDataAccess.afterInsert()
```
beikov: I checked and there is no `NonStrictReadWriteEntityDataAccess`, only `EntityNonStrictReadWriteAccess`, even in Hibernate ORM 5.3, so I'm a bit curious what the LLM hallucinated here 😅 

```suggestion
  at EntityNonStrictReadWriteAccess.afterInsert()
```

The `afterInsert` method doesn't do anything for that class, so what assertion error did you encounter?
I'm asking because an `AssertionError` would be a bug in our code. Can you please share the real and full stack trace?

posts/Stale/2026-07-01-hibernate-immutable-2lc-9x-speedup.adoc
```diff
@@ -0,0 +1,359 @@
...
+----
+
+`NONSTRICT_READ_WRITE` assumes entities have a `@Version` field for optimistic locking. Our entity doesn't have one, it doesn't need one, because it's immutable. 
+The assertion failure occurs when the strategy tries to verify the version after inserting into the cache.
```
beikov: I couldn't find anything, javadoc, inline comments or documentation, that supports these claims. Please share stack traces and further details.

posts/Stale/2026-08-19-hibernate-immutable-2lc-9x-speedup.adoc
```diff
@@ -0,0 +1,359 @@
...
+Marking an entity `@Immutable` means Hibernate will not generate UPDATE statements for it. But our pipeline has two cases where value data legitimately changes:
+
+1. *Deduplication*: when a recalculation produces a different result for an existing value
+2. *Ephemeral nullification*: nulling out intermediate computation data to save storage
+
+Both are handled with native SQL plus explicit cache eviction:
+
+[source,java]
+----
+// Mutation via native SQL (bypasses @Immutable restriction)
```
beikov, line 260: You can surely use native queries if you prefer that, but note that you can also configure `hibernate.query.immutable_entity_update_query_handling_mode=allow` and then use HQL updates as well.

posts/Stale/2026-07-01-hibernate-immutable-2lc-9x-speedup.adoc
beikov: This makes no sense, since reference caching is disabled when the entity type contains a collection. Also see `AbstractEntityPersister#shouldUseReferenceCacheEntries`.

posts/Stale/2026-07-01-hibernate-immutable-2lc-9x-speedup.adoc
```diff
@@ -0,0 +1,359 @@
...
+[source,java]
+----
+@ManyToMany(cascade = { CascadeType.PERSIST }, fetch = FetchType.EAGER)
+@JoinTable(name = "value_edge", ...)
+public List<ValueEntity> sources;
+----
+
+With the L2 cache, cached entities can be returned outside the original Hibernate session. Accessing a lazy collection on a cached entity throws `LazyInitializationException` because there is no session to initialize the proxy.
+
+Changing to `FetchType.EAGER` means the sources collection is loaded with the entity and cached together. Since subsequent loads hit the L2 cache, the eager fetch cost is paid only once per entity.
```
beikov: Can you please try to annotate the collection with `@Immutable @Cache(usage = CacheConcurrencyStrategy.READ_ONLY)` instead of marking it as `EAGER`?

posts/Stale/2026-07-01-hibernate-immutable-2lc-9x-speedup.adoc
```diff
@@ -0,0 +1,359 @@
...
+
+* 500 uploads (214,000 values): stable 340 values/second throughput
+* 1,327 uploads (308,000 values): no degradation
+
+== Key Takeaways
+
+*Give Hibernate accurate information about your data model.* The difference between "cacheable entity" and "fully annotated immutable entity" was 9.5x. But this requires annotations at multiple levels: `@Immutable` on the entity (controls dirty-checking), `@Cache(READ_ONLY)` on the entity (controls cache entry strategy), AND `@Mutability(Immutability.class)` on complex fields (controls the type system's deep copy and serialization). Missing any one of these leaves significant performance on the table. In our case, omitting the field-level annotation alone caused a 7.6x slowdown because the JSON type's default `MutabilityPlan` performs full serialize/parse round-trips on every cache operation.
+
+*`@Immutable` and `@Cache(READ_ONLY)` are designed to work together.* Using `@Immutable` with a read-write cache strategy is a mismatch, it either fails (with `@Version` assertion errors) or wastes resources on unnecessary concurrency controls. `READ_ONLY` is the only strategy that fully leverages immutability.
```
beikov: Again, please share the details about these errors. It seems to me that this might be the LLM hallucinating.

posts/Stale/2026-07-01-hibernate-immutable-2lc-9x-speedup.adoc
```diff
@@ -0,0 +1,359 @@
...
+
+*Native SQL queries bypass the L2 cache.* This is easy to miss. If your performance-critical queries use native SQL (as ours do for recursive CTEs), the L2 cache provides zero benefit unless you restructure queries to fetch IDs only, then load entities via `em.find()`.
```
beikov: Again, maybe mention the use of CTEs with HQL as an alternative solution.

posts/Stale/2026-07-01-hibernate-immutable-2lc-9x-speedup.adoc
gavinking: To be clear, `hibernate.cache.use_reference_entries` is something we see as quite a sharp knife and it's not something that casual visitors should be playing about with.

posts/Stale/2026-07-01-hibernate-immutable-2lc-9x-speedup.adoc
stalep: I profiled with async-profiler (JFR mode, Hibernate 7.4.x via Quarkus 3.35.1) using @Immutable + @Cache(READ_ONLY) but without @Mutability(Immutability.class) on the JSONB field. Removing @Mutability alone causes an 8.1x regression (90s → 733s for 20 imports producing ~10.5k values).
The deepCopy call chain when loading the entity via a JPQL query:
```
EntityInitializerImpl.deepCopy(EntityPersister, Object[], Object[]) line: 878
  EntityInitializerImpl.takeSnapshot(...) line: 1705
    EntityInitializerImpl.initializeEntityInstance(...) line: 1626
      EntityInitializerImpl.initializeInstance(...) line: 1501
        StandardRowReader.coordinateInitializers() line: 239
          StandardRowReader.readRow(...) line: 137
            ... (SqmQueryImpl.doList())
```
The field-level deepCopy goes through:
```
MutableMutabilityPlan.deepCopy()
  SerializableJavaType$SerializableMutabilityPlan.deepCopyNotNull()
    SerializationHelper.clone()
      ObjectOutputStream.writeObject() / ObjectInputStream.readObject()
```
Our JqValue type implements Serializable, so Hibernate resolves SerializableJavaType as the JavaType and its MutabilityPlan does full Java object serialization for every deepCopy. Out of ~540k CPU samples, ~267k (49%) were in deepCopy.
With `@Mutability(Immutability.class)` the deepCopy becomes an identity function and the cost disappears entirely.
I agree with your intuition that deepCopy should not be needed for @Immutable entities, if takeSnapshot could skip or use an identity copy for immutable entities, the `@Mutability` annotation would not be needed as a workaround. I have the full JFR file available if you want, but I can't attach it here.

posts/Stale/2026-08-19-hibernate-immutable-2lc-9x-speedup.adoc
stalep: Ah, I did not know that, just confirmed and that works, thanks!

posts/Stale/2026-07-01-hibernate-immutable-2lc-9x-speedup.adoc
stalep: You're right, I could not reproduce that error. I  tested` @Immutable` + `@Cache(NONSTRICT_READ_WRITE)` + `@Mutability(Immutability.class)` and it works fine with no assertion errors (all tests pass, benchmark completes at 30.1s vs 25.0s for READ_ONLY). I'll stick with READ_ONLY as it's faster though.

posts/Stale/2026-08-19-hibernate-immutable-2lc-9x-speedup.adoc
stalep: I did not know about that, thanks!

posts/Stale/2026-07-01-hibernate-immutable-2lc-9x-speedup.adoc
stalep: Yes, our entity has a `@ManyToMany` collection, so `shouldUseReferenceCacheEntries` returns false regardless of the cache strategy or `hibernate.cache.use_reference_entries` setting. Reference caching was never used in our case. The post incorrectly attributed the speedup to ReferenceCacheEntryImp, I'll update it.

posts/Stale/2026-07-01-hibernate-immutable-2lc-9x-speedup.adoc
stalep: Yes, so since reference caching is disabled for our entity (it has a collection). The `LazyInitException` claim in the blog was wrong. I need EAGER fetch because our app access the collection outside of session/transaction boundaries. I'll update this.

posts/Stale/2026-07-01-hibernate-immutable-2lc-9x-speedup.adoc
stalep: https://github.com/hibernate/in.relation.to/pull/269#discussion_r3508949852

posts/Stale/2026-07-01-hibernate-immutable-2lc-9x-speedup.adoc
stalep: Did not know, thanks!

posts/Stale/2026-08-19-hibernate-immutable-2lc-9x-speedup.adoc
stalep: I was not aware of that either, this will require a bit more work/testing on my side, but I'll try to test it by today.

posts/Stale/2026-07-01-hibernate-immutable-2lc-9x-speedup.adoc
stalep: I tested that, switched the sources collection to LAZY with `@Immutable @Cache(READ_ONLY)` cause a lot of failures (NPEs from sources being null).

posts/Stale/2026-07-01-hibernate-immutable-2lc-9x-speedup.adoc
beikov: When the entity is marked as `@Immutable`, the `EntityPersister` should report `false` for `isMutable()` and hence the snapshotting shouldn't happen as you can see in the code: https://github.com/hibernate/hibernate-orm/blob/7.4/hibernate-core/src/main/java/org/hibernate/sql/results/graph/entity/internal/EntityInitializerImpl.java#L1842

Can you share that branch for further investigation please? I'm curious what the exact conditions are that trigger this.

stalep: It's from h5m main; https://github.com/Hyperfoil/h5m look at the `ValueEntity`

posts/Stale/2026-08-19-hibernate-immutable-2lc-9x-speedup.adoc
gavinking: I wonder if I messed up when I changed that setting to `ERROR`-by-default in H7.

Perhaps I should have just removed it. 

I only changed it to error because it used to default to `WARNING`, which was garbage. But perhaps I should have changed it to `ALLOW` instead. 

The issue which introduced this thing (HHH-12387) doesn't provide any strong reason for it to even exist. I wonder what @sebersole thinks.

posts/Stale/2026-07-01-hibernate-immutable-2lc-9x-speedup.adoc
stalep: I dug deeper and my earlier stack trace was misleading. You were right that `takeSnapshot` skips `deepCopy` for `@Immutable` entities.
I confirmed isMutable() returns false for our entity (Hibernate 7.3.2 via Quarkus 3.35.1). I also re-profiled and looked at the actual sample distribution more carefully. The 2 samples showing EntityInitializerImpl.takeSnapshot → deepCopy out of 267k total were noise. The real cost is in the 2LC cache assembly path:
```
CacheLoadHelper.loadFromSecondLevelCache()
  CacheLoadHelper.convertCacheEntryToEntity()       (~11k samples)
    StandardCacheEntryImpl.assemble()                (~10k samples)
      CacheEntryHelper.assemble()
        AbstractStandardBasicType.assemble()
          MutableMutabilityPlan.assemble()
            SerializableJavaType$SerializableMutabilityPlan.deepCopyNotNull()
              SerializationHelper.clone()             ← full Java serialization
```
Plus the cache put path:
```
StandardCacheEntryImpl.<init>()                      (~1k samples)
  CacheEntryHelper.disassemble()
    AbstractStandardBasicType.disassemble()
      MutableMutabilityPlan.disassemble()
        SerializableJavaType$SerializableMutabilityPlan.deepCopyNotNull()
```
So the deepCopy/serialization cost is from StandardCacheEntryImpl's "by value" semantics on every cache get/put, not from dirty-checking snapshots. Since our JqValue type implements Serializable, Hibernate uses SerializableJavaType whose MutabilityPlan does SerializationHelper.clone() (full ObjectOutputStream/ObjectInputStream round-trip) for assemble, disassemble, and deepCopy.
@Mutability(Immutability.class) fixes this because it replaces the field's MutabilityPlan with identity operations: assemble/disassemble/deepCopy all return the same reference. I think.... :)

posts/Stale/2026-07-01-hibernate-immutable-2lc-9x-speedup.adoc
```diff
@@ -0,0 +1,364 @@
...
+concurrent read-write access patterns. In our benchmarks, `READ_ONLY` was approximately 20% faster than `NONSTRICT_READ_WRITE` for the same workload 
+(25.0s vs 30.1s for 5 imports producing ~2,600 values). Both strategies work correctly with `@Immutable` entities.
+
+=== @Mutability(Immutability.class) on the JSONB field
+
+This turned out to be *critical*, and its importance was not obvious. We initially assumed `@Immutable` on the entity would be sufficient, 
+but benchmarking revealed an 8.1x performance difference from this single field annotation.
+
+The reason: `@Immutable` controls *entity-level* dirty-checking, but Hibernate's type system has its own *field-level* `MutabilityPlan` that operates independently. 
+Even on an `@Immutable` entity, the field-level `deepCopy` is still called during entity initialization. We confirmed this with async-profiler (JFR mode),
```
beikov: It would be nice if you mention that this happens, because values have to be disassembled to be stored in cache and assembled when retrieved from cache, which is what invokes the `deepCopy` method, also see the `disassemble`/`assemble` method implementations in `SerializableJavaType`, inherited from `MutableMutabilityPlan`. This has nothing to do with entity initialization per-se, so this section is a bit misleading IMO.

posts/Stale/2026-07-01-hibernate-immutable-2lc-9x-speedup.adoc
```diff
@@ -0,0 +1,364 @@
...
+
+| `@Immutable` + `@Cache(READ_ONLY)` without `@Mutability`
+| *733s*
+|===
+
+*The entity-level `@Immutable` is not enough.* The field-level `MutabilityPlan` must also be told that the value is immutable, 
+especially for types implementing `Serializable` where the default plan performs full Java object serialization round-trips.
+
+NOTE: It is worth investigating whether `EntityInitializerImpl.takeSnapshot()` could skip `deepCopy` entirely for `@Immutable` entities, 
+since dirty-checking is not performed for them. This could eliminate the need for the `@Mutability` workaround.
```
beikov: This is obsolete, now that we figured that the copying comes from the assembling/disassembling of cache values.

```suggestion
```

posts/Stale/2026-07-01-hibernate-immutable-2lc-9x-speedup.adoc
stalep: I've played with that for a couple of hours now and I did get several of the CTEs to work with HQL. However when I ran `EXPLAIN ANALYZE` on the new hql queries they are not as performant with a higher execution and planning time unfortunately.

posts/Stale/2026-07-01-hibernate-immutable-2lc-9x-speedup.adoc
```diff
@@ -0,0 +1,364 @@
...
+AND `@Mutability(Immutability.class)` on complex fields (controls the type system's deep copy and serialization). 
+Missing any one of these leaves significant performance on the table. In our case, omitting the field-level annotation alone caused an 8.1x slowdown 
+because the default `MutabilityPlan` for `Serializable` types performs full Java object serialization round-trips on every `deepCopy` call, 
+even on `@Immutable` entities.
+
+*Entity-level and field-level immutability are independent.* `@Immutable` on the entity class controls dirty-checking (should Hibernate generate UPDATEs?). 
+`@Mutability(Immutability.class)` on a field controls the type system's `MutabilityPlan` (how should the value be copied and serialized?). 
+For scalar fields like `int` or `String`, the default `MutabilityPlan` is cheap. For types implementing `Serializable` (including custom JSON types), 
+the default plan performs full Java object serialization, and this cost applies even on `@Immutable` entities because `EntityInitializerImpl.takeSnapshot()` 
+still calls `deepCopy` during entity initialization from query results.
```
beikov: Again, it should be clarified, that this is due to cache value assembling/disassembling, not for snapshotting.

posts/Stale/2026-07-01-hibernate-immutable-2lc-9x-speedup.adoc
beikov: Alright, that makes a lot more sense. Good thing you double checked. Now you just need to update the sections to clarify this :)

beikov (approved):

posts/Stale/2026-07-01-hibernate-immutable-2lc-9x-speedup.adoc
```diff
@@ -0,0 +1,367 @@
...
+
+*Result: no measurable improvement.* The benchmark ran at essentially the same speed 285.9 seconds before, 285.7 seconds after.
+
+We enabled the cache, Hibernate was storing entities in it, the hit rate looked reasonable, but the wall clock didn't move. Something was wrong, and profiling would tell us what.
+
+=== What the profiler revealed
+
+async-profiler CPU flamegraphs showed the top consumers:
+
+[cols="3,1,1"]
```
marko-bekhta, line 54:
```suggestion
[cols="3,1,1", stripes="even", options="header"]
```
Could you also use ^ for the tables ? It should produce a slightly more readable table 🙂 
(I think there are a few more in this post)

yrodiere (approved):

yrodiere:
Thanks. The build failure was probably because we updated some dependencies on the target branch. I force-pushed to try and fix that. Let's merge if the build passes!

yrodiere:
> Thanks. The build failure was probably because we updated some dependencies on the target branch. I force-pushed to try and fix that. Let's merge if the build passes!

It was actually some obscure Asciidoc syntax that we don't handle well. I added a commit to work around that.

PR #272, opened 2026-07-21
Add workflows to preview the PRs

marko-bekhta:


.github/workflows/ci.yml
```diff
@@ -0,0 +1,30 @@
...
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # 7.0.0
+
+      - name: Build site
+        run: |
+          docker run --pull always --rm -t -u "$(id -u):$(id -g)" -v "$WORKSPACE_DIR":/home/dev/website:rw quay.io/hibernate/awestruct-build-env:latest "rake setup && rake clean[all] gen[production]"
```
yrodiere: FWIW I think there's a `[dev]` profile to only build the last 6 months of blog posts or so. If you have problems with the build being slow, you can try that.

marko-bekhta: nice 🙂 thanks
I was about to say that the build seems quite fast... but then I tried the dev profile and ... well 😁 it's waaay faster 🙂

yrodiere (approved):
I guess we need to merge to see the workflow in action? If so, feel free to go ahead.

Note there is also a Jenkinsfile, which maybe we should remove as I think your GHA workflow pretty much covers everything we do in the Jenkinsfile. Also there's probably some MAINTAINERS.md to update or something...

PR #274, opened 2026-08-19
Fix titles containing colon-space being mangled by YAML.load

yrodiere-agent:
## Summary

- Awestruct's `AsciidoctorHandler` passes all parsed header values through `YAML.load`. When a title contains `: ` (colon-space), YAML interprets it as a mapping and returns a Hash instead of a String — breaking layouts that expect `page.title` to be a String.
- Added a `fix_title` method to the `Posts` extension that detects when `page.title` has been mangled into a Hash and reconstructs the original title string.
- Removed the double-quote workaround (`= "Title: Subtitle"`) from 16 existing posts that no longer need it.

## Test plan

- [ ] Run `rake clean gen` and verify the site builds without errors
- [ ] Spot-check a few posts with `: ` in the title (e.g. "Release: Hibernate ORM 5.4.22.Final", "Hibernate Reactive: Beta 2") to verify titles render correctly
- [ ] Verify existing posts whose titles were already quoted still render correctly

### DIFF
--- a/_ext/posts.rb
+++ b/_ext/posts.rb
@@ -6,6 +6,27 @@ module Awestruct
 
       attr_accessor :path_prefix, :assign_to, :archive_template, :archive_path, :default_layout, :wp_compat
 
+      # Awestruct's AsciidoctorHandler passes all parsed header values through
+      # YAML.load (see asciidoctor_handler.rb, parse_document_attributes).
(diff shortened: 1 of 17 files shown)


yrodiere (approved):
Build passed, rendering seems fine, I'm happy. Merging.

